PR #303, opened 2017-04-12
adding 'safeguard' to catch and erorr an ErrorObservable thereby not breaking the chain

WrathOfZombies:
@Zlatkovsky could you test the changes locally once before merging this. I couldn't test it thoroughly although I did test the concept.


Zlatkovsky:
@cshashwat , have you already fixed the observer issue with a pull request of your own?  I think I remember seeing something like this when I was on vacation.  If so, shall we go ahead and close this one?

cshashwat:
This may reproduce at other points in the code (have only added my fix where we knew there were issues) but know how to fix this issue now. Can close this pull request.

PR #372, opened 2017-04-21
Disable AJAX tracking (app insights) + non-code (readme, tutorial) changes

Zlatkovsky:


Zlatkovsky:
@saunders77 , @WrathOfZombies , I'd like to merge today's "master" into insiders (and then prod)
There is only one actual code-change here, a one-liner:

    this._current.config.disableAjaxTracking = true;

The rest are all non-code files (readme + tutorial file, which now gets auto-open)

Let me know if I should go ahead and push to insiders, and then to prod.

Issue #375, opened 2017-04-25
Unexpected error when deactivating Firebug


maninweb:
Hi there,

I searched, if something similar has already been posted here, but could not find a topic. I hope, I did not oversee something. When deactivating Firebug, I get an unexpected error from the Add-In. Please see the screenshot below:

![image](https://cloud.githubusercontent.com/assets/13045898/25394154/c1da9434-29dd-11e7-921c-8d11882f2a17.png)

Also, when clicking then on "Load the last snippet", nothing happens (or may be, the error is reloaded). The other buttons work. For reproducing, please follow these steps...

- Open the task panes for Code and Run from the Script Lab ribbon.
- Create a new blank snippet.
- Run the blank snippet.
- Please click on 'Deactivate Firebug for this page'.
- Change something in the code window, e.g. just add a line.
- The runner task pane asks you to refresh.
- Please refresh and the error appears.

Sometimes, after going back to the page with the list of snippets for running the last snippet again and repeating the steps for producing the error page, the error page did not show the "Back" button. However I could not reproduce this constantly.

My current version of Script Lab is 1.1.0 (Deployed 2017-04-16 21:53 pm UTC) and my User ID is cj1taawts0000586wmychyiyx.

Best,
Mourad

Zlatkovsky:
Thanks for the detailed report!  I will try it out later this week.  Just to be sure, based on the screenshots,  this is on a Windows build of Office, correct?

maninweb:
Hi, yes - I had the error in Excel 2016 Version 1704 Build 8067.2018 on a Windows 10 and Windows 7 OS.

Zlatkovsky:
OK, perfect, thanks for the clarification.  Will take a look in the coming days.

Just to check:  if you re-open the panes, does Firebug come back?  Or are you essentially in a blocked / broken state now, and cannot proceed to use Script Lab?  This will also help gage the severity of this issue.

maninweb:
If I get the error and go back to the list of scripts inside the Run Task Pane by clicking on the arrow (left) and then call the script again, Firebug automatically reappears. If I close the Task Pane and reopen it from the ribbon, I get automatically see this list of scripts and Firebug also appears again. When having the Run Task Pane open with the error inside and then clicking in the ribbon on Run, nothing happens.
[...]

PR #381, opened 2017-04-26
Remove a few more extraneous FirebugLite buttons

Zlatkovsky:
Namely, remove detach-firebug (since it causes runtime error on heartbeat refresh; see https://github.com/OfficeDev/script-lab/issues/375)
Also removing the detach to different window (since not sure how useful that is in practice for snippets in the stripped-down version of firebug, without dom explorer, etc.), and removing the firebug logo-button as it offers a bunch of options that may be impractical and/or cause issues in our case.
Since there now is only a single red minimize button available, making it graysacle to make it stand out less.

### DIFF
--- a/src/client/assets/firebug/firebug-lite-debug.js
+++ b/src/client/assets/firebug/firebug-lite-debug.js
@@ -9890,48 +9890,53 @@ append(ChromeBase,
             [
                 {
                     label: "Open Firebug",
-                    type: "shortcut",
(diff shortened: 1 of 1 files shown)


Zlatkovsky:
On further testing, looks like the experience isn't perfect in Internet Explorer.  Namely:

- Icon isn't gray (though it is in Chrome)
- Strange issue with button

image

Zlatkovsky:
Fixing with a follow-up commit soon, but not ready to merge in yet.

PR #391, opened 2017-04-28
README details on using external libraries + some reorganization

Zlatkovsky:
I added *a bunch of detail* on how to use external libraries, and also used this as an opportunity for some minor reorganization.

Could one of you take a look and see if it all makes sense and is grammatically sound?  @WrathOfZombies, could you also help make sure I'm using the right terminology and explanation of the various types of modules?  Thanks!


WrathOfZombies:
Would you want to split this readme into Wiki's instead? It would be easier to consume and only look for what you want thanks to a TOC.

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+
+    https://unpkg.com/spectrum-colorpicker@1.8.0/spectrum.js
+    https://unpkg.com/spectrum-colorpicker@1.8.0/spectrum.css
+
+> Tip: To see the available files in a package, simply include a `/` after the package name, and you'll get a directory listing.  E.g., <https://unpkg.com/spectrum-colorpicker/>. It will also conveniently tell you the available version numbers, via a dropdown at the top right of the site:
+>
+> ![Import](.github/images/unpkg.jpg)
+
+
+#### Moderate: NPM package is an AMD module, but the library is also available via a "dist" build
```
WrathOfZombies: Wrong title for the given example. An AMD module is one that is loaded from require.js

WrathOfZombies: dont mention "dist" in title as its not a hard and fast rule to have stuff available from dist.

markpete:
Is it worth talking about nested library dependencies and how they get loaded?  My assumption is that this is all automatic like a normal npm package, but I'm curious if there are cases where you need to add the nested dependencies to the Library list.

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+     "https://unpkg.com/javascript-qrcode",
+     0, 0, undefined]
+
+In Internet Explorer (and hence on Add-ins on a PC version of Office), you don't get much more information than this -- but on a different platform (or Office Online on a different browser), you will see:
+
+    ["Uncaught ReferenceError: module is not defined",
+     "https://unpkg.com/javascript-qrcode@1.0.7",
+     3302, 1, ReferenceError: module is not defined {}]
+
+If you are lucky, even though the *NPM* package doesn't contain a web-ready version of the package, the web-compatible version may be available elsewhere.  In the case of this `javascript-qrcode` package, the [NPM entry](https://www.npmjs.com/package/javascript-qrcode) points to a GitHub repo, <https://github.com/siciarek/javascript-qrcode>. And the repo, in turn, has a "dist" folder (convention), along with exampe usage that also shows `<script src="dist/qrcode.min.js"></script>`.
```
WrathOfZombies: Instead taking a fail first approach would help, as the developer to verify that the library is web compatible. Then try and reference via unpkg. If it still doesnt work, examine the contents of the package, try rawgit. else post an issue on the developers repository to expose an npm package.

README.md
WrathOfZombies: Based on your further docs, i think you mean to add a NOT?

`NPM package is an **NOT** AMD module, b`

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+    });
+
+If you're lucky, it will work right off the bat.  If you're slightly less lucky, you might still get a useful error that will help you fix the issue (which is why the `require.onError` line described in step #1 is so important!)
+
+    Error: Script error for "tinycolor", needed by: tinygradient
+    http://requirejs.org/docs/errors.html#scripterror {...}
+
+This just means that you need to have required an additional library, tinycolor.  Again, find its location (with https://unpkg.com being the easiest way of procuring it), and add it to the config paths list (*again, remembering to omit just the `.js` part of the filename). Also include the library name in the array in the `require` statement.  When you patch up both bits, you will end up with the following:
+
+    // At the top of the file:
```
WrathOfZombies: mark code range with ```typescript instead of tabs/4 spaces.. and end it with ```
This will allow you to get coloring.

siewmoi:
Wikis are great but I found wiki not so portable (compared to md files).   When I was doing MS Graph samples we used shared wiki to point ppl to small how to this and that.   Keeping track of them was a pain too.

markpete:
Also, what version(s) of ES does ScriptLab support?  I'm assuming ES5/6 yes and ES7 no.  Is it the same as Edge, IE, or Chrome?  Chakra?

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+Include this in your Libraries tab, and you're good to go!
+
+
+#### Harder: NPM package is an AMD module (must be loaded with `require.js`), and is *not* available in "dist" form
+
+Suppose you get the same error as in the previous section, but you can't find a dist version of a library that you want.  Let's take [tinygradient](https://github.com/mistic100/tinygradient) as one such example.  The GitHub documentation of tinygradient says that it is "Compatible with Require.js/AMD and NodeJS".  Let's use Require.js to load it.
+
+In your libraries tab, add a reference to `require.js` and to its IntelliSense types (note that `require.js` itself should be loaded from a web-compatible CDN version, not its NPM package):
+
+    https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.3/require.min.js
```
WrathOfZombies: use `unpkg!!!` 

```
requirejs
@types/requirejs
```

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+Almost there!  To be consumed inside of Script Lab, we need the file to be served in *RAW* form, ***and*** with correct Content-Type headers.  To do this, use a service like [RawGit](https://rawgit.com) to transform the file for you, and give you a URL you can use directly.
+
+In this case, pasting in the above URL into RawGit produces:
+
+    https://cdn.rawgit.com/siciarek/javascript-qrcode/041ad555/dist/qrcode.min.js
+
+Include this in your Libraries tab, and you're good to go!
+
+
+#### Harder: NPM package is an AMD module (must be loaded with `require.js`), and is *not* available in "dist" form
```
WrathOfZombies: Technically this is the `medium` difficulty one. Given that there's a possibility of loading stuff in.

@Zlatkovsky a todo would be adding a require.config tab when we see require.js being a dependency.

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+    });
+
+3. Add the code that you want to use, *inside of a `require` block*:
+
+    require(['tinygradient'], function (tinygradient) {
+        var gradient = tinygradient('red', 'blue');
+        console.log(gradient.rgb(10));
+    });
+
+If you're lucky, it will work right off the bat.  If you're slightly less lucky, you might still get a useful error that will help you fix the issue (which is why the `require.onError` line described in step #1 is so important!)
```
WrathOfZombies: Something about this feels weird. Luck has nothing to do with it. 
Its code. there's either a true or a false.

README.md
```diff
@@ -64,14 +74,173 @@ If you have a suggestion for a feature, please feel free to file it under "issue
...
+
+    Error: Script error for "tinycolor", needed by: tinygradient
+    http://requirejs.org/docs/errors.html#scripterror {...}
+
+This just means that you need to have required an additional library, tinycolor.  Again, find its location (with https://unpkg.com being the easiest way of procuring it), and add it to the config paths list (*again, remembering to omit just the `.js` part of the filename). Also include the library name in the array in the `require` statement.  When you patch up both bits, you will end up with the following:
```
WrathOfZombies: I feel we are teaching folks how to use require.js today. In that case how do we plan to support say system.js or X.js when its out tomorrow. May be we should avoid this. Have it as a blog post on your personal blog or medium saying "How do you load AMD libraries in Script Lab". There you can explain all you want.

WrathOfZombies:
@markpete  we support upto ES7 and we down compile back to ES5.

WrathOfZombies:
@siewmoi Could you explain what you meant by port? Where are we porting them to? Also why would we port something on this repository elsewhere?

PR #415, opened 2017-05-08
Export to zip file for publishing

Zlatkovsky:
Add ability to export to a zip file

image

Resulting zip file looks like this (here, shown already extracted)

image

And can produce a running website:

image


Zlatkovsky:
@WrathOfZombies , I will go ahead and merge this in, since I think it adds a lot of value, and seems to work quite well.  There are places, esp. wrt. observables in GitHub effects, where the code could probably be improved, so would love your feedback -- but I don't think it's blocking deploying this.

PR #421, opened 2017-05-13
Fix language display

Zlatkovsky:
Hook up language display at bottom of editor

Today, language display was always empty (null).  Looks like the issues is that the language-related actions were taking two parameters rather than just the single payload parameter.  Fixed by unifying payload to accept complex objects like

{ type, language }

### DIFF
--- a/src/client/app/actions/monaco.ts
+++ b/src/client/app/actions/monaco.ts
@@ -27,7 +27,7 @@ export class MonacoActionTypes {
 export class ChangeTabAction implements Action {
     readonly type = MonacoActionTypes.CHANGE_TAB;
 
-    constructor(public payload: { type: string, language: string }) {}
+    constructor(public payload: { name: string, language: string }) {}
 }
 
 export class ResetAction implements Action {
--- a/src/client/app/actions/monaco.ts
+++ b/src/client/app/actions/monaco.ts
@@ -27,7 +27,7 @@ export class MonacoActionTypes {
 export class ChangeTabAction implements Action {
     readonly type = MonacoActionTypes.CHANGE_TAB;
 
-    constructor(public payload: string, public language: string) { }
+    constructor(public payload: { type: string, language: string }) {}
 }
 
 export class ResetAction implements Action {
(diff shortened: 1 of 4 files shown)


src/client/app/actions/monaco.ts
```diff
@@ -27,7 +27,7 @@ export class MonacoActionTypes {
 export class ChangeTabAction implements Action {
     readonly type = MonacoActionTypes.CHANGE_TAB;
 
-    constructor(public payload: string, public language: string) { }
+    constructor(public payload: { name: string, language: string }) {}
```
WrathOfZombies: we should be sending multiple parameters when possible. Is there a reason to this?

src/client/app/actions/monaco.ts
```diff
@@ -39,13 +39,13 @@ export class ResetAction implements Action {
 export class UpdateIntellisenseAction implements Action {
     readonly type = MonacoActionTypes.UPDATE_INTELLISENSE;
 
-    constructor(public payload: string[], public language: string) { }
+    constructor(public payload: { libraries: string[], language: string }) {}
```
WrathOfZombies: we should be sending multiple parameters when possible. Is there a reason to this?

src/client/app/actions/monaco.ts
```diff
@@ -39,13 +39,13 @@ export class ResetAction implements Action {
...
 }
 
 export class AddIntellisenseAction implements Action {
     readonly type = MonacoActionTypes.ADD_INTELLISENSE;
 
-    constructor(public payload: string, public language: string) { }
+    constructor(public payload: { url: string, language: string }) { }
```
WrathOfZombies: we should be sending multiple parameters when possible. Is there a reason to this?

src/client/app/containers/editor.ts
```diff
@@ -43,14 +43,23 @@ export class Editor implements AfterViewInit {
...
+    }
 
     updateIntellisense() {
         if (this.snippet == null) {
             return;
         }
 
-        this._store.dispatch(new Monaco.UpdateIntellisenseAction(this.snippet.libraries.split('\n'), 'typescript'));
+        this._store.dispatch(new Monaco.UpdateIntellisenseAction(
+            { libraries: this.snippet.libraries.split('\n'), language: 'typescript' }
```
WrathOfZombies: we should be sending multiple parameters when possible. Is there a reason to this?

src/client/app/actions/monaco.ts
Zlatkovsky: I don't know why, but when the constructor was declared as payload and then another parameter, the other parameter was getting lost on the receiving end.  It seems as though it wasn't getting serialized, or something like that.  I then noticed that all of your other constructors always had a single parameter to them -- hence I merged it in to a single object, and things started working again!  Without it, language was always null

Zlatkovsky: @WrathOfZombies, let me go ahead and merge it in as is.  Ultimately I don't think it matters (it works in my version, and didn't in the original, so it's still a net gain; and you have some other places where `payload` takes in an object anyway).  If you want to go ahead and explore it for academic curiosity, by all means!  But I think I may as well merge it in for the time being.

PR #440, opened 2017-06-02
adding api search code for libraries

WrathOfZombies:
Still needs work but provides the initial basic intellisense. falls back to the current behavior if the service is unresponsive.


WrathOfZombies:
@Zlatkovsky do you want @cshashwat to take a look at this?

Zlatkovsky:
I do, I haven't forgotten about this one.  Though priority / order-of-operations-wise, there are a number of other things I'd like him to do first.

Issue #27, opened 2017-01-10
Ability to update shared Gists


Zlatkovsky:

Zlatkovsky:
* If shared as a gist, probably ok to just update in place.   (Maybe add a "duplicate" functioanilty for the odd case)
* If import an existing gist (via URL):
  * if it's your gist (check), then resume using it as is and updating as is
  * if it's someone else's, clear out the "gist" field and assume it's new.
* [Optional: re-parenting option to re-specify (and view) current gist ID].  Maybe under snippet profile, should have option to input a different gist id if you please.

PR #453, opened 2017-06-20
Added ability to update GitHub gists

cshashwat:
Added ability to update GitHub gists

## Description

A user can now update gists in GitHub, if he/she owns the gist. Previously, every time a user clicked "Public new gist" or "Private new gist" under the "Share" menu, a new copy was created. The proposed feature would allow users to update their gists in place, with GitHub showing each new version.

If a user does not own the gist (e.g. importing the gist from another person's account), he/she will not be able to update the gist. However upon sharing as a public or private gist, the user now has permission and can update his/her own copy of the gist.

The "Share" menu is updated to reflect the new "Update Existing Gist" option; strings have been updated for clarity for preexisting options as well. If a user imports a gist, then deletes it from GitHub and tries to click "Update" from Script Lab, the app will trigger a full refresh, similar to the existing import failure. Publishing a gist, public or private, does not create new local copies of that gist, but updates the gist's ID instead with the new one provided by GitHub.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/27

## Motivation and Context

See above link to issue. Users may or may not want additional copies of gists in their accounts; they now have the functionality to choose.

## How Has This Been Tested?

Testing was done by following the existing scenarios in the new file 'TESTING.md' under the 'Share' section, with other scenarios added to reflect the new proposed feature.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


WrathOfZombies:
@cshashwat Will take a look at it by tomorrow EOD. Thanks for the contribution 👍

TESTING.md
```diff
@@ -0,0 +1,22 @@
...
+  * Running deleted snippet has reasonable behavior.
+  * [Side-by-side runner]:
+    * Edit to code causes runner want to refresh.
+    * Whether starting from error or going to error state and back out, should act correctly.
+  * Note: when running a script on localhost, you may see two compiler warnings for a critical dependency in the console log. These two warnings always appear in debug mode and are normal.
+* Sharing:
+  * Copying to clipboard works
+  * Can share as gist, public and private
+  * Can update existing gist
+  * Deleting a gist and then trying to update it via local copy should prompt full refresh, after which update option should be gone
```
dsaunders45459: Should mention the case that the "Update existing Gist" button becomes available immediately after initial publish of a new Gist.

TESTING.md
```diff
@@ -0,0 +1,22 @@
...
+  * Erroneous code (e.g, syntax error) shows error correctly.
+  * Running deleted snippet has reasonable behavior.
+  * [Side-by-side runner]:
+    * Edit to code causes runner want to refresh.
+    * Whether starting from error or going to error state and back out, should act correctly.
+  * Note: when running a script on localhost, you may see two compiler warnings for a critical dependency in the console log. These two warnings always appear in debug mode and are normal.
+* Sharing:
+  * Copying to clipboard works
+  * Can share as gist, public and private
+  * Can update existing gist
```
dsaunders45459: Should also mention to test the case where you import a Gist that you do not own and try to update it.

src/client/app/containers/app.ts
```diff
@@ -144,7 +145,7 @@ export class AppComponent {
         this._store.dispatch(new UI.ToggleImportAction(true));
     }
 
-    shareGist(isPublic: boolean) {
+    shareGist(isPublic: boolean, update: boolean) {
```
dsaunders45459: The project should follow and document a uniform naming convention. It looks like booleans should be prefixed with is* for this function.

src/client/app/effects/snippet.ts
```diff
@@ -144,6 +144,29 @@ export class SnippetEffects {
...
+        .ofType(Snippet.SnippetActionTypes.UPDATE_INFO)
+        .map(( { payload } ) => {
+            let { id, name, description, gist, owned } = payload;
+            let snippet: ISnippet = storage.lastOpened;
+            if (storage.snippets.contains(id)) {
+                snippet = storage.snippets.get(id);
+
+                /* check if fields are undefined or null */
+                name != null ? snippet.name = name : snippet.name = snippet.name;
+                description != null ? description = payload.description : snippet.description = snippet.description;
```
dsaunders45459: I am not sure why we assign to description and then no longer use the local variable.

src/client/app/effects/snippet.ts
```diff
@@ -144,6 +144,29 @@ export class SnippetEffects {
...
+    updateInfo$: Observable<Action> = this.actions$
+        .ofType(Snippet.SnippetActionTypes.UPDATE_INFO)
+        .map(( { payload } ) => {
+            let { id, name, description, gist, owned } = payload;
+            let snippet: ISnippet = storage.lastOpened;
+            if (storage.snippets.contains(id)) {
+                snippet = storage.snippets.get(id);
+
+                /* check if fields are undefined or null */
+                name != null ? snippet.name = name : snippet.name = snippet.name;
```
dsaunders45459: snippet.name = snippet.name looks like a complicated no-op. Can we just change all of these
if (name != null) {
    snippet.name = name;
}

src/client/app/reducers/github.ts
```diff
@@ -48,7 +48,8 @@ export function reducer(state = initialState, action: GitHubActions): GitHubStat
         }
 
         case GitHubActionTypes.SHARE_PRIVATE_GIST:
-        case GitHubActionTypes.SHARE_PUBLIC_GIST: {
+        case GitHubActionTypes.SHARE_PUBLIC_GIST:
+        case GitHubActionTypes.UPDATE_GIST: {
             AI.trackEvent(action.type);
             return { ...state, sharing: true };
```
dsaunders45459: Can we log a hashed gist id?

src/client/app/services/github.service.ts
```diff
@@ -95,7 +95,7 @@ export class GitHubService {
         if (this.profile == null) {
             return Observable.of([]);
         }
-        let url = `${this._baseUrl}/users/${this.profile.login}/gists`;
+        let url = `${this._baseUrl}/users/${this.profile.login}/gists?timestamp=${(new Date).getTime()}`;
```
dsaunders45459: It looks like angular doesn't support cache busting by default like jQuery, but you could also add  'Cache-Control' : 'no-cache' to the headers to prevent the browser cache.

src/interfaces/playground.d.ts
```diff
@@ -6,6 +6,7 @@ interface ITemplate {
     ///////////////////////////////////////////////////////////////////////////////////////////////////
     id?: string;
     gist?: string;
+    owned?: boolean;
```
dsaunders45459: We should follow standard naming conventions throughout the project. I think this should be named isOwned

TESTING.md
cshashwat: Added.

TESTING.md
cshashwat: Added.

src/client/app/containers/app.ts

src/client/app/effects/snippet.ts
cshashwat: This was a typo on my part, thanks for catching it. Fixed this and modified the ternary operators to be if statements instead, since the 'snippet.field = snippet.field' is an unnecessary no-op.

src/client/app/effects/snippet.ts
cshashwat: Added; see above comment.

src/client/app/reducers/github.ts
cshashwat: Added under UPDATE_INFO in 'snippet.ts' in 'reducers'; it logs the gist id using the same sha hash function.

src/client/app/services/github.service.ts
cshashwat: Discussed offline. Added optional parameter for get() function in 'app/services/request.ts' to use cache or not; if this is set to 'false', then the timestamp random number param is applied. The GitHub API seems to reject using 'Cache-Control' in the headers. Would welcome other suggestions.

src/interfaces/playground.d.ts
cshashwat: Changed throughout code to 'isOwned'

src/client/app/reducers/snippet.ts
```diff
@@ -81,6 +82,16 @@ export function reducer(state = initialState, action: SnippetActions | GitHubAct
         case SnippetActionTypes.STORE_UPDATED:
             return { ...state, loading: false };
 
+        case SnippetActionTypes.UPDATE_INFO: {
+            let updatedInfo = { };
+            if (action.payload.gist != null) {
+                updatedInfo['gistHashedId'] = sha1(action.payload.gist.toString());
```
dsaunders45459: Is gist already a string? If it is there is no need to call toString on it.

src/client/app/helpers/snippet.helper.ts
```diff
@@ -72,7 +73,6 @@ export const snippetFieldSortingOrder: { [key: string]: number } = {
 export function getSnippetDefaults(): ISnippet {
     return {
         id: '',
-        gist: '',
```
Zlatkovsky: Why did you remove the empty-string gist?  It seemed to be a reasonable default...

src/client/app/helpers/strings.ts
```diff
@@ -52,7 +53,8 @@ export class Strings {
     static readonly githubLogoutFailed = 'Failed to sign out of GitHub';
     static readonly profileCheckFailed = 'Failed to get the GitHub profile';
     static readonly gistRetrieveFailed = 'Failed to retrieve GitHub gists';
-    static readonly gistShareFailed = 'Failed to share the GitHub gist';
+    static readonly gistShareFailedBody = 'Failed to share the GitHub gist';
+    static readonly gistShareFailedTitle = 'Share failed';
```
Zlatkovsky: I'd probably do "Sharing failed" instead of "Share failed".
In general, we've been doing string peer-review by putting `/* NEEDS STRING REVIEW */` next to new strings, and having Michael Saunders then look through them.  Search for this phrase and see other instances where I have it

src/client/app/services/github.service.ts
```diff
@@ -96,7 +96,7 @@ export class GitHubService {
             return Observable.of([]);
         }
         let url = `${this._baseUrl}/users/${this.profile.login}/gists`;
-        return this._request.get<IGist[]>(url, ResponseTypes.JSON, this._headers);
+        return this._request.get<IGist[]>(url, ResponseTypes.JSON, this._headers, false);
```
Zlatkovsky: Add a parameter comment:

    return this._request.get<IGist[]>(url, ResponseTypes.JSON, this._headers, false /*useCache*/);

src/client/app/services/request.ts
```diff
@@ -19,7 +19,11 @@ export class Request {
         return this._response(xhr, responseType);
     }
 
-    get<T>(url: string, responseType: ResponseTypes, headers?: Object): Observable<T> {
+    get<T>(url: string, responseType: ResponseTypes, headers?: Object, useCache?: boolean): Observable<T> {
```
Zlatkovsky: generally an optional boolean parameter should be something where the default is `false`.  Here you have it as true.
Could you reverse the naming/functionality such that it's more of a `forceBypassCache`?

src/client/app/services/request.ts
```diff
@@ -19,7 +19,11 @@ export class Request {
...
+        if (useCache != null && !useCache) {
+            url += (url.indexOf('?') === -1) ? '?' : '&';
```
Zlatkovsky: How about a single statement?

            url = url + ((url.indexOf('?') === -1) ? '?' : '&') + `timestamp=${(new Date).getTime()}`;

src/client/app/actions/snippet.ts
```diff
@@ -65,7 +65,7 @@ export class ImportSuccessAction implements Action {
 export class UpdateInfoAction implements Action {
     readonly type = SnippetActionTypes.UPDATE_INFO;
 
-    constructor(public payload: { name: string, description: string }) { }
+    constructor(public payload: { id: string, name?: string, description?: string, gist?: string, isOwned?: boolean }) { }
```
cshashwat: Discussed offline. This has been addressed by the field `gistOwnerId` in the `IGist` interface. The check is done in the actual app.

src/client/app/containers/app.ts
```diff
@@ -18,8 +18,9 @@ import { Strings, environment } from '../helpers';
                     <command icon="OpenPaneMirrored" title="${Strings.runSideBySide}" (click)="runSideBySide()"></command>
                 </command>
                 <command [hidden]="isEmpty" icon="Share" [async]="sharing$|async" title="${Strings.share}">
-                    <command icon="PageCheckedin" title="${Strings.shareMenuPublic}" (click)="shareGist(true)"></command>
-                    <command icon="ProtectedDocument" title="${Strings.shareMenuPrivate}" (click)="shareGist(false)"></command>
+                    <command *ngIf="snippet?.isOwned" icon="Save" title="${Strings.updateMenu}" (click)="shareGist(null, true)"></command>
```
cshashwat: Haven't been able to find a way to leave inline comments in the Angular template. Perhaps HTML comments, or there's something I've missed? Would welcome any suggestions.

src/client/app/effects/github.ts
```diff
@@ -84,10 +84,13 @@ export class GitHubEffects {
 
     @Effect()
     shareGist$: Observable<Action> = this.actions$
-        .ofType(GitHub.GitHubActionTypes.SHARE_PRIVATE_GIST, GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST)
+        .ofType(GitHub.GitHubActionTypes.SHARE_PRIVATE_GIST,
+        GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST,
+        GitHub.GitHubActionTypes.UPDATE_GIST)
```
cshashwat: Fixed.

src/client/app/effects/github.ts
```diff
@@ -98,10 +101,14 @@ export class GitHubEffects {
             description.replace(Strings.gistDescriptionAppendage, ''); // shouldn't be necessary
             description += Strings.gistDescriptionAppendage;
 
+            if (type === GitHub.GitHubActionTypes.UPDATE_GIST && payload.isOwned) {
+                id = payload.gist;
```
cshashwat: Fixed.

src/client/app/effects/github.ts
```diff
@@ -120,14 +127,21 @@ ${Strings.gistSharedDialogEnd}
...
-        .mergeMap(gist => Observable.from([
-            new GitHub.LoadGistsAction(),
-            new GitHub.ShareSuccessAction(gist)
-        ]))
-        .catch(exception => Observable.from([
-            new UI.ReportErrorAction(Strings.gistShareFailed, exception),
-            new GitHub.ShareFailedAction()
-        ]));
+        .mergeMap((gist) => Observable.from([
+                new GitHub.LoadGistsAction(),
```
cshashwat: Fixed.

src/client/app/effects/github.ts
```diff
@@ -120,14 +127,21 @@ ${Strings.gistSharedDialogEnd}
...
+        )
+        .catch(exception => {
+            this._uiEffects.alert(
+                Strings.gistShareFailedBody + '\n\n' + Strings.reloadPrompt,
+                Strings.gistShareFailedTitle,
+                Strings.okButtonLabel)
+            .then(() => window.location.reload());
+            return Observable.from([
+                new GitHub.ShareFailedAction(),
+                new Snippet.UpdateInfoAction({ id: storage.lastOpened.id, gist: '', isOwned: false })]);
```
cshashwat: Thanks for bringing this up. Have modified code to pass through the `id` from the `payload` up above. That should address concerns with using `storage` (which has now been removed) and the un-owning of the snippet (since the `isOwned` field has now also been removed).

src/client/app/effects/snippet.ts
```diff
@@ -144,6 +144,37 @@ export class SnippetEffects {
...
+    updateInfo$: Observable<Action> = this.actions$
+        .ofType(Snippet.SnippetActionTypes.UPDATE_INFO)
+        .map(( { payload } ) => {
+            let { id, name, description, gist, isOwned } = payload;
+            let snippet: ISnippet = storage.lastOpened;
+            if (storage.snippets.contains(id)) {
+                snippet = storage.snippets.get(id);
+
+                /* check if fields are undefined or null */
+                if (name != null) {
```
cshashwat: Discussed offline. Have fixed this by adding the `isNil()` from lodash.

src/client/app/effects/snippet.ts
```diff
@@ -144,6 +144,37 @@ export class SnippetEffects {
...
+                    snippet.name = name;
+                }
+                if (description != null) {
```
cshashwat: Addressed above.

src/client/app/effects/snippet.ts
```diff
@@ -144,6 +144,37 @@ export class SnippetEffects {
...
+                if (gist != null) {
+                    snippet.gist = gist;
+                }
+                if (isOwned != null) {
+                    snippet.isOwned = isOwned;
+                }
+
+                storage.snippets.insert(id, snippet);
+            }
+            return snippet;
```
cshashwat: Fixed.

src/client/app/helpers/snippet.helper.ts
```diff
@@ -29,6 +29,7 @@ const snippetFields: { [key: string]: SnippetFieldType } = {
     /* ITemplate base class */
     id: SnippetFieldType.INTERNAL,
     gist: SnippetFieldType.INTERNAL,
+    isOwned: SnippetFieldType.INTERNAL,
```
cshashwat: Have addressed this by adding your suggestion. Thanks.

src/client/app/helpers/snippet.helper.ts
cshashwat: Had forgotten to revert this change I'd made while I was testing out some code. Thanks for pointing it out. Fixed.

src/client/app/helpers/strings.ts
cshashwat: Modified strings and added `/* NEEDS STRING REVIEW */` for all strings I've added to the file.

src/client/app/containers/app.ts
```diff
@@ -64,6 +66,18 @@ export class AppComponent {
         return /commands=1/ig.test(location.search);
     }
 
+    get isOwned() {
```
Zlatkovsky: I still think we want a better name for it.

src/client/app/containers/app.ts
```diff
@@ -18,8 +19,9 @@ import { Strings, environment } from '../helpers';
                     <command icon="OpenPaneMirrored" title="${Strings.runSideBySide}" (click)="runSideBySide()"></command>
                 </command>
                 <command [hidden]="isEmpty" icon="Share" [async]="sharing$|async" title="${Strings.share}">
-                    <command icon="PageCheckedin" title="${Strings.shareMenuPublic}" (click)="shareGist(true)"></command>
-                    <command icon="ProtectedDocument" title="${Strings.shareMenuPrivate}" (click)="shareGist(false)"></command>
+                    <command *ngIf="isOwned|async" icon="Save" title="${Strings.updateMenu}" (click)="shareGist(false, true)"></command>
```
Zlatkovsky: please add comments to boolean (or any other non-obvious) paramaters.

e.g.,

    shareGist(false /*parameter1Name*/, true /*parameter2Name*/)

src/client/app/effects/github.ts
```diff
@@ -98,14 +103,20 @@ export class GitHubEffects {
             description.replace(Strings.gistDescriptionAppendage, ''); // shouldn't be necessary
             description += Strings.gistDescriptionAppendage;
 
+            if (type === GitHub.GitHubActionTypes.UPDATE_GIST) {
+                gistId = gist;
+            }
+
             return this._github.createOrUpdateGist(
                 `${description}`,
```
Zlatkovsky: isn't that just the same as `description`?

src/client/app/effects/github.ts
```diff
@@ -98,14 +103,20 @@ export class GitHubEffects {
...
                 files,
-                null,
-                type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
+                gistId,
+                type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST).map((gist: IGist) => {
```
Zlatkovsky: could we make this whole thing read cleaner?

    let isPublicGist = type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
    return this._github
        .createOrUpdateGist(
            description, files, gistId, isPublicGist
        )
        .map(gist: IGist => ({ gist: gist, snippetId: id} ));

src/client/app/effects/github.ts
```diff
@@ -98,14 +103,20 @@ export class GitHubEffects {
...
+                    return { gist: gist, snippetId: id };
```
Zlatkovsky: if you wrap it in parentheses (see my prettified code above), you don't need the separate

    {
        return { <something> };
    }

PR #457, opened 2017-06-26
Small UI changes for Gists

cshashwat:
Added 'Gist URL' field to snippet info box, also added confirmation boxes for sharing public and private gists

## Description

In snippet info box, accessed by clicking on the snippet's name, there is now a 'Gist URL' field, in addition to the preexisting 'Name' and 'Description' fields. This field is only populated if the Gist has been imported or has already been shared: the URL displayed is its GitHub URL and is grayed out and cannot be modified. For newly created/non-shared snippets, this field will not show.

Added confirmation boxes for sharing public and private gists. Now user will be prompted before sharing snippet as gist.

## Motivation and Context

Designed to aid users with gist workflow: can now see Gist URLs (if they exist), which helps them make sure they're working on correct gist. And prompting before sharing ensures that users understand they are creating a new copy of theirs gist.

## How Has This Been Tested?

Testing has been done by the updated changes in `TESTING.md`.

Screenshots of changes follow:

Sharing Gist confirmation:

capture

Gist URL in snippet info:

capture

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.

### DIFF
--- a/src/client/app/components/snippet.info.ts
+++ b/src/client/app/components/snippet.info.ts
@@ -19,8 +19,8 @@ import { isNil } from 'lodash';
                 </div>
 
                 <div *ngIf="!!url" class="ms-TextField">
-                    <label class="ms-Label">${Strings.urlLabel}</label>
-                    <input class="ms-TextField-field" type="text" [(ngModel)]="url" placeholder="" disabled />
+                    <label class="ms-Label">${Strings.gistUrlLabel}</label>
+                    <a href="{{url}}" target="_blank">Open in browser</a>
                 </div>
             </div>
             <div class="ms-Dialog-actions">
--- a/src/client/app/components/snippet.info.ts
+++ b/src/client/app/components/snippet.info.ts
@@ -16,6 +16,11 @@ import { Strings } from '../helpers';
                     <label class="ms-Label">${Strings.descriptionLabel}</label>
                     <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="snippet.description" placeholder="Description of the snippet"></textarea>
                 </div>
+
+                <div *ngIf="url" class="ms-TextField ms-TextField--multiline">
+                    <label class="ms-Label">${Strings.urlLabel}</label>
+                    <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="url" disabled="true"></textarea>
+                </div>
             </div>
             <div class="ms-Dialog-actions">
                 <div class="ms-Dialog-actionsRight">
@@ -35,4 +40,11 @@ export class SnippetInfo {
     @Input() snippet: ISnippet;
     @Input() show: boolean;
     @Output() dismiss = new EventEmitter<ISnippet>();
+
+    get url() {
+        if (!this.snippet.gist) {
+            return null;
+        }
+        return `https://gist.github.com/${this.snippet.gist}`;
+    }
 }
--- a/src/client/app/components/snippet.info.ts
+++ b/src/client/app/components/snippet.info.ts
@@ -1,5 +1,6 @@
 import { Component, Input, ChangeDetectionStrategy, Output, EventEmitter } from '@angular/core';
 import { Strings } from '../helpers';
+import { isNil } from 'lodash';
 
 @Component({
     changeDetection: ChangeDetectionStrategy.OnPush,
@@ -42,9 +43,6 @@ export class SnippetInfo {
     @Output() dismiss = new EventEmitter<ISnippet>();
 
     get url() {
-        if (!this.snippet.gist) {
-            return null;
-        }
-        return `https://gist.github.com/${this.snippet.gist}`;
+        return isNil(this.snippet.gist) ? null : `https://gist.github.com/${this.snippet.gist}`;
     }
 }
--- a/src/client/app/components/snippet.info.ts
+++ b/src/client/app/components/snippet.info.ts
@@ -18,9 +18,9 @@ import { isNil } from 'lodash';
                     <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="snippet.description" placeholder="Description of the snippet"></textarea>
                 </div>
 
-                <div *ngIf="url" class="ms-TextField ms-TextField--multiline">
+                <div *ngIf="!!url" class="ms-TextField">
                     <label class="ms-Label">${Strings.urlLabel}</label>
-                    <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="url" disabled="true"></textarea>
+                    <input class="ms-TextField-field" type="text" [(ngModel)]="url" placeholder="" disabled />
                 </div>
             </div>
             <div class="ms-Dialog-actions">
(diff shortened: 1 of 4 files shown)


src/client/app/components/snippet.info.ts
```diff
@@ -16,6 +17,11 @@ import { Strings } from '../helpers';
                     <label class="ms-Label">${Strings.descriptionLabel}</label>
                     <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="snippet.description" placeholder="Description of the snippet"></textarea>
                 </div>
+
+                <div *ngIf="url" class="ms-TextField ms-TextField--multiline">
+                    <label class="ms-Label">${Strings.urlLabel}</label>
+                    <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="url" disabled="true"></textarea>
```
dsaunders45459: Should use <input type="text"> for a single line field.

src/client/app/components/snippet.info.ts
```diff
@@ -16,6 +17,11 @@ import { Strings } from '../helpers';
                     <label class="ms-Label">${Strings.descriptionLabel}</label>
                     <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="snippet.description" placeholder="Description of the snippet"></textarea>
                 </div>
+
+                <div *ngIf="url" class="ms-TextField ms-TextField--multiline">
```
dsaunders45459: !!url is more standard for a null check.

src/client/app/containers/app.ts
```diff
@@ -175,10 +175,26 @@ export class AppComponent {
                     this._store.dispatch(new GitHub.UpdateGistAction(this.snippet));
                 }
                 else if (isPublic) {
-                    this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
+                    let sharePublicGist = async() => {
+                        let result = await this._effects.alert(Strings.sharePublicSnippetConfirm, `${Strings.share} ${this.snippet.name}`, `${Strings.share}`, `${Strings.cancelButtonLabel}`);
```
dsaunders45459: Strings.share is already a string, don't think you need to embed it into another string such as `${Strings.share}`

src/client/app/containers/app.ts
```diff
@@ -175,10 +175,26 @@ export class AppComponent {
...
+                        return result;
```
dsaunders45459: I think you can just return the result promise, no need to await for it here as we do not do anything with the return value in this function.

src/client/app/components/snippet.info.ts
Zlatkovsky: For the gist URL in the about snippet dialog, should we just do "Gist ID:  Open in browser" with a hyperlink?

src/client/app/containers/app.ts
Zlatkovsky: Agreed, you don't need to do template strings around Strings.share and Strings.cancelButtonLabel

src/client/app/components/snippet.info.ts
cshashwat: Field is now single-line.

src/client/app/components/snippet.info.ts
cshashwat: Added.

src/client/app/containers/app.ts
cshashwat: Fixed.

src/client/app/containers/app.ts
cshashwat: Fixed.

src/client/app/containers/app.ts
```diff
@@ -175,10 +175,20 @@ export class AppComponent {
...
+                    let sharePublicGist = () => (this._effects.alert(Strings.sharePublicSnippetConfirm, `${Strings.share} ${this.snippet.name}`, Strings.share, Strings.cancelButtonLabel));
+                    sharePublicGist().then((result: string) => {
+                        if (result !== Strings.cancelButtonLabel) {
+                            this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
+                        }
+                    });
                 }
                 else {
-                    this._store.dispatch(new GitHub.SharePrivateGistAction(this.snippet));
+                    let sharePrivateGist = () => (this._effects.alert(Strings.sharePrivateSnippetConfirm, `${Strings.share} ${this.snippet.name}`, Strings.share, Strings.cancelButtonLabel));
```
Zlatkovsky: Why are you assigning this to a variable, rather than using directly?

src/client/app/helpers/strings.ts
```diff
@@ -82,6 +82,9 @@ export class Strings {
     static readonly nameLabel = 'Name';
     static readonly descriptionLabel = 'Description';
 
+    /** NEEDS STRING REVIEW */
+    static readonly urlLabel = 'Gist URL';
```
Zlatkovsky: how about gistUrlLabel?

src/client/app/components/snippet.info.ts
```diff
@@ -16,6 +17,11 @@ import { Strings } from '../helpers';
                     <label class="ms-Label">${Strings.descriptionLabel}</label>
                     <textarea class="ms-TextField-field ms-font-m" [(ngModel)]="snippet.description" placeholder="Description of the snippet"></textarea>
                 </div>
+
+                <div *ngIf="!!url" class="ms-TextField">
```
Zlatkovsky: What if we instead had it as a link:

     Gist URL:  <a href="..." target="_blank">Open in browser</a>

cshashwat: Added.

src/client/app/components/snippet.info.ts
```diff
@@ -19,8 +19,8 @@ import { isNil } from 'lodash';
                 </div>
 
                 <div *ngIf="!!url" class="ms-TextField">
-                    <label class="ms-Label">${Strings.urlLabel}</label>
-                    <input class="ms-TextField-field" type="text" [(ngModel)]="url" placeholder="" disabled />
+                    <label class="ms-Label">${Strings.gistUrlLabel}</label>
+                    <a href="{{url}}" target="_blank">Open in browser</a>
```
Zlatkovsky: This should be a string from the `Strings.` class.  Will merge in as soon as you address this.

src/client/app/containers/app.ts
```diff
@@ -175,20 +175,22 @@ export class AppComponent {
                     this._store.dispatch(new GitHub.UpdateGistAction(this.snippet));
                 }
                 else if (isPublic) {
-                    let sharePublicGist = () => (this._effects.alert(Strings.sharePublicSnippetConfirm, `${Strings.share} ${this.snippet.name}`, Strings.share, Strings.cancelButtonLabel));
-                    sharePublicGist().then((result: string) => {
-                        if (result !== Strings.cancelButtonLabel) {
-                            this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
+                    this._effects.alert(Strings.sharePublicSnippetConfirm, `${Strings.share} ${this.snippet.name}`, Strings.share, Strings.cancelButtonLabel)
+                    .then((result: string) => {
```
Zlatkovsky: I think you want an extra tabbing to the `.then` potion.  Ditto below.

src/client/app/components/snippet.info.ts
cshashwat: Added.

src/client/app/containers/app.ts

Issue #401, opened 2017-05-01
Add ctrl+tab (and visual contextual action) for IntelliSense


Zlatkovsky:
See combo of 

    editor.getAction('editor.action.triggerSuggest');

and <https://microsoft.github.io/monaco-editor/playground.html#interacting-with-the-editor-adding-an-action-to-an-editor-instance>

May want to ask is there is a way of extending shortcut / duplicating action.

Zlatkovsky:
Similar to https://github.com/Microsoft/monaco-editor/issues/225

PR #463, opened 2017-06-29
Added trigger suggest hotkey and context menu item

kevishi:
## Description

Added a Trigger Suggest hotkey (F2) and context menu item.

## Testing

Hotkey and context item tested in Excel, Chrome, Edge, Firefox.

### DIFF
--- a/src/client/app/containers/editor.ts
+++ b/src/client/app/containers/editor.ts
@@ -38,8 +38,7 @@ export class Editor implements AfterViewInit {
      * Initialize the component and subscribe to all the necessary actions.
      */
     async ngAfterViewInit() {
-        this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
-        let editor = this._monacoEditor;
+        const editor = await this._monaco.create(this._editor, { theme: 'vs' });
         editor.addAction({
             id: Strings.triggerActionId,
             label: Strings.triggerActionLabel,
@@ -50,8 +49,10 @@ export class Editor implements AfterViewInit {
             contextMenuOrder: 0,
             run: () => editor.trigger('ngAfterViewInit', Strings.monacoTriggerAction, {})
         });
+        this._monacoEditor = editor;
         this._createTabs();
         this._subscribeToState();
+
     }
 
     changeTab = (name: string = 'script') => {
--- a/src/client/app/containers/editor.ts
+++ b/src/client/app/containers/editor.ts
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
      */
     async ngAfterViewInit() {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let editor = this._monacoEditor;
+        editor.addAction({
+            id: 'trigger-suggest',
+            label: 'Trigger Suggest',
+            keybindings: [monaco.KeyCode.F2],
+            keybindingContext: null,
+            contextMenuGroupId: 'navigation',
+            contextMenuOrder: 5.5,
+            run: () => editor.trigger('ngAfterViewInit', 'editor.action.triggerSuggest', {})
+        });
         this._createTabs();
         this._subscribeToState();
     }
--- a/src/client/app/containers/editor.ts
+++ b/src/client/app/containers/editor.ts
@@ -41,6 +41,7 @@ export class Editor implements AfterViewInit {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
         let editor = this._monacoEditor;
         editor.addAction({
+            // Unique id for action
             id: 'trigger-suggest',
             label: Strings.triggerActionLabel,
             keybindings: [monaco.KeyCode.F2],
--- a/src/client/app/containers/editor.ts
+++ b/src/client/app/containers/editor.ts
@@ -46,7 +46,8 @@ export class Editor implements AfterViewInit {
             keybindings: [monaco.KeyCode.F2],
             keybindingContext: null,
             contextMenuGroupId: 'navigation',
-            contextMenuOrder: 0,
+            // Put action at top of context menu
+            contextMenuOrder: 0, 
             run: () => editor.trigger('ngAfterViewInit', 'editor.action.triggerSuggest', {})
         });
         this._createTabs();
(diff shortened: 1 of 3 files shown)


src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
      */
     async ngAfterViewInit() {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let editor = this._monacoEditor;
```
Zlatkovsky: Why are you creating a new variable for this?  why not just re-use `this._monacoEditor` as is, and call `addAction` on it.

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
...
+        editor.addAction({
+            id: 'trigger-suggest',
+            label: 'Trigger Suggest',
+            keybindings: [monaco.KeyCode.F2],
+            keybindingContext: null,
+            contextMenuGroupId: 'navigation',
+            contextMenuOrder: 5.5,
```
Zlatkovsky: Could you add a comment for what `5.5` means?

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
      */
     async ngAfterViewInit() {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let editor = this._monacoEditor;
+        editor.addAction({
+            id: 'trigger-suggest',
```
Zlatkovsky: Where is this ID used?

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
...
+            label: 'Trigger Suggest',
```
Zlatkovsky: Please add it to the Strings file

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,16 @@ export class Editor implements AfterViewInit {
...
+            keybindings: [monaco.KeyCode.F2],
+            keybindingContext: null,
+            contextMenuGroupId: 'navigation',
+            contextMenuOrder: 5.5,
+            run: () => editor.trigger('ngAfterViewInit', 'editor.action.triggerSuggest', {})
+        });
```
Zlatkovsky: What about removing some of the unneeded (top) items in the context menu?

src/client/app/containers/editor.ts
kevishi: I need to access it in the lambda function in addAction's run. Inside the function defintion "this" becomes "undefined"

src/client/app/containers/editor.ts
kevishi: The id is internal to monaco, it needs to be "unique" to show up.

src/client/app/containers/editor.ts
Zlatkovsky: Oh, I see, interesting.  That might be worth pointing out in a comment.  And in that case I would initially do a `const editor = await this._monaco.create(...)`, and then set `this._monacoEditor = editor` at the very end.

src/client/app/containers/editor.ts
Zlatkovsky: ok, please add a comment to this effect.

src/client/app/containers/editor.ts
kevishi: From my brief dive into context menu options this seems to be hidden in monaco internals

src/client/app/helpers/strings.ts
```diff
@@ -112,6 +112,11 @@ export class Strings {
 
     static readonly deleteSnippetConfirm = 'Are you sure you want to delete this snippet?';
 
+    static readonly triggerActionId = 'trigger-suggest';
```
Zlatkovsky: No no.  Please only include actual localizable strings in here (only the `triggerActionLabel`).  The rest should be constants in `editor.ts`, but it will never be localized or even make sense form here, so no need to be here in Strings.

kevishi: that makes much more sense

src/client/app/helpers/strings.ts
```diff
@@ -112,6 +112,8 @@ export class Strings {
 
     static readonly deleteSnippetConfirm = 'Are you sure you want to delete this snippet?';
 
+    static readonly triggerActionLabel = 'Trigger Suggest';
```
Zlatkovsky: When you think of this getting localized across multiple languages and people needing to figure out what this string means:  should it be more like `editorTriggerSuggestLabel` or `editorTriggerSuggestMenuItemLabel`?

Other than that, looks good.  Please push one final commit, and then I will merge it in.

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,17 @@ export class Editor implements AfterViewInit {
...
     async ngAfterViewInit() {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let editor = this._monacoEditor;
+        editor.addAction({
+            id: 'trigger-suggest',
+            label: Strings.triggerActionLabel,
+            keybindings: [monaco.KeyCode.F2],
+            keybindingContext: null,
+            contextMenuGroupId: 'navigation',
+            // Put action at top of context menu
```
Zlatkovsky: Could you make it be a `/* ... */` sort of comment?  I.e.,:

        contextMenuOrder: 0 /* put at top of context menu */,

src/client/app/containers/editor.ts
```diff
@@ -39,6 +39,17 @@ export class Editor implements AfterViewInit {
      */
     async ngAfterViewInit() {
         this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let editor = this._monacoEditor;
+        editor.addAction({
+            id: 'trigger-suggest',
```
Zlatkovsky: could you add a comment:

        id: 'trigger-suggest' /* must be unique id */

src/client/app/helpers/strings.ts
kevishi: Now that I actually remember what it does I would think either 'editorTriggerSuggestMenuItemLabel' or 'editorTriggerSuggestContextMenuLabel'

Zlatkovsky:
Resolves https://github.com/OfficeDev/script-lab/issues/401

PR #464, opened 2017-06-29
Environment change

cshashwat:
Added ability to switch environments based on drop down menu in "About" tab. Also added fixes for Office UI fabric rendering issues on Safari

## Description

Users can now change their environments between:
- Production
- Edge
- Insiders

by way of a drop down menu in the "About" tab. Selecting an environment different from the one the user is currently in will first prompt the user to confirm his/her choice; afterwards, the app will reload in the new environment, including the runner panes, etc.

Once the user has switched from `localhost` to another environment, the user should clear the browser's local storage and cache for the `localhost` domain to switch back; other environments do not have the `localhost` option on the drop down.

## Motivation and Context

This change is intended to allow users to dogfood new Script Lab features/fixes in different environments.

## How Has This Been Tested?

Tested via sideloading local manifest into Office and into browser, to ensure that redirect worked.

Dropdown menu:

capture5

User confirmation before switching, after clicking OK button on 'About' dialog:

capture6

## Types of changes

## Checklist:


Zlatkovsky:
UI comment:  it feels like the switcher is placed somewhat randomly in the about screen. 
Should we instead create an "environment" category (visually, in terms of spacing), and then have `current: https://localhost:3000`, followed by `switch: <dropdown>`.  Nothing super fancy, just reorganizing the same information to make the environment feel like a single entity rather than placed in the middle of the screen and not logically grouped to the label above it.

image

config/env.config.js
```diff
@@ -55,5 +55,25 @@ const config = {
...
+            compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
+                let htmlHead = htmlPluginData.html.match('<head>').index;
+                htmlPluginData.html = htmlPluginData.html.slice(0, htmlHead) + '<head>' +
+                    `<script>
+                        var newUrl = window.localStorage.getItem('environmentConfig');
+                        if (newUrl && !window.location.href.match(newUrl)) {
+                            window.location.href = newUrl + window.location.pathname;
+                        }
+                    </script>`
+                    + htmlPluginData.html.slice(htmlHead + 6);
```
Zlatkovsky: Nit, code style: could you place the ` + ` on the line above instead?

Zlatkovsky: I'm assuming 6 is the length of `<head>`?  Could you either put it as an inlined comment next to `6`, or just use `'<head>'.length` directly?

    htmlPluginData.html.slice(htmlHead + 6 /* '<head>' */);

or, better IMO

    htmlPluginData.html.slice(htmlHead + '<head>'.length);

config/env.config.js
```diff
@@ -55,5 +55,25 @@ const config = {
     }
 };
 
+const redirect = function(options) {
```
Zlatkovsky: I feel like `redirect` could use being more descriptive (out of context, would you think it's a verb? a noun?..).

Maybe `redirectionAppender` or `redirectionAdder`?  ... OK, neither sounds great either, but I'm sure there's something clever you can come up with :-)

config/webpack.common.js
```diff
@@ -155,6 +155,11 @@ module.exports = (prodMode) =>
                 filename: 'heartbeat.html',
                 template: './views/heartbeat.html',
                 chunks: ['polyfills', 'vendor', 'heartbeat'],
-            })
+            }),
+            new HtmlWebpackPlugin({
+                filename: 'tutorial.html',
+                template: './views/tutorial.html',
+            }),
+            new redirect({ options: '' })
```
Zlatkovsky: What are these options used for?  I don't see you use them above, either?

manifests/script-lab-edge.xml
```diff
@@ -17,7 +17,11 @@
     <AppDomain>https://github.com</AppDomain>
     <AppDomain>https://stackoverflow.com</AppDomain>
     <AppDomain>https://dev.office.com</AppDomain>
-    <AppDomain>https://bornholm-runner-edge.azurewebsites.net</AppDomain>  
+    <AppDomain>https://script-lab.azureedge.net</AppDomain>
+    <AppDomain>https://script-lab-runner.azureedge.net</AppDomain>
+    <AppDomain>https://bornholm-runner-edge.azurewebsites.net</AppDomain>
```
Zlatkovsky: Could you order these in some order like production => insiders => edge?
I would also be OK with all environments having all the same app domains list.  Yeah, it'd be slightly overkill, but it's easier to copy-paste, and it wouldn't harm anything to have an extra app domain (even if it's already implied based on where the page is running from)

src/client/app/components/about.ts
```diff
@@ -29,9 +37,28 @@ import { environment, storageSize, Strings, storage } from '../helpers';
...
+        private _changeDetector: ChangeDetectorRef
+    ) { }
+
+    ngAfterViewInit() {
+        if (environment.current.config.name === config['local'].name) {
+            this.configs.push({ name: 'local', value: 'Localhost' });
+        }
+        this.configs.push(
+            { name: 'production', value: 'Production' },
+            { name: 'edge', value: 'Beta' },
```
Zlatkovsky: Um... that's the reverse.

Production = production
insiders = beta (still fairly baked)
edge = gets latest from master, is the least stable.  hence alpha

src/client/app/components/about.ts
```diff
@@ -29,9 +37,28 @@ import { environment, storageSize, Strings, storage } from '../helpers';
...
     @Output() showChange = new EventEmitter<boolean>();
+
+    constructor(
+        private _effects: UIEffects,
+        private _changeDetector: ChangeDetectorRef
+    ) { }
+
+    ngAfterViewInit() {
+        if (environment.current.config.name === config['local'].name) {
+            this.configs.push({ name: 'local', value: 'Localhost' });
```
Zlatkovsky: Nit: Something about the capitalization of `Localhost` just looks weird.  What if, just for this one, we would write out its url (`config.editorUrl`).  I.e., `https://localhost:3000`.  But use the `editorUrl` variable, it's safer in case the ports ever change.  Or just `https://localhost`, that's fine too.

src/client/app/components/about.ts
```diff
@@ -42,4 +69,30 @@ export class About {
     ${storageSize(localStorage, `playground_${environment.current.host}_snippets`, Strings.aboutSnippets)}
     ${storageSize(sessionStorage, 'playground_intellisense', Strings.aboutIntellisense)}
     `;
+
+    configs = [];
```
Zlatkovsky: I prefer variable initializations at the top of the file, above the constructor.  Otherwise it's hard to trace through where these things are declared.

src/client/app/helpers/strings.ts
```diff
@@ -115,6 +118,7 @@ export class Strings {
     /** NEEDS STRING REVIEW */
     static readonly sharePublicSnippetConfirm = 'Are you sure you want to share this snippet as a new public gist?';
     static readonly sharePrivateSnippetConfirm = 'Are you sure you want to share this snippet as a new secret gist?';
+    static readonly changeEnvironmentConfirm = 'You are about to change your Script Lab environment and will not have access to your saved snippets until you return to this environment. Are you sure you want to proceed?';
```
Zlatkovsky: Good string!  But I would replace "saved snippets" with "saved local snippets", to clarify the difference about local workspaces vs. things saved as gists

config/env.config.js
cshashwat: Added as `RedirectPlugin`.

config/env.config.js
cshashwat: Fixed.

cshashwat: Added.

config/webpack.common.js
cshashwat: It's unnecessary. Addressed and replaced with an empty object.

manifests/script-lab-edge.xml
cshashwat: Added the same AppDomains to every manifest, with the exception of `localhost` and its runner domain. Do we want to have `localhost` be an option for the user to switch back to or no?

src/client/app/components/about.ts
cshashwat: Added option of `https://localhost:3000` to drop down menu.

src/client/app/components/about.ts
cshashwat: Thanks for clarifying this. Fixed.

src/client/app/components/about.ts
cshashwat: Fixed.

src/client/app/helpers/strings.ts
cshashwat: Fixed.

cshashwat:
Updated screenshots to reflect new UI changes.

WrathOfZombies:
@cshashwat Could you follow the UI to use standard design patterns? The dropdown looks weird. Use the fabric dropdown instead. You will need to write an ng2 directive around it.

cshashwat:
@WrathOfZombies I tried using the Fabric JS dropdown component earlier. The issue was that the Angular2 bindings, such as `[(ngModel)]` and `[value]` were not working. They don't have a ng2 directive for Office Fabric JS for the dropdown, as far as I can see:

https://github.com/ngOfficeUIFabric/ng-officeuifabric
https://github.com/ngOfficeUIFabric/ng-officeuifabric/issues/405

I then tried manually listing the options and adding the `selected` attribute to the `option` tag for the current environment, even though this is messier. That didn't work. They claim that they support it here:

https://github.com/OfficeDev/office-ui-fabric-core/issues/321

But this is using JQuery, and a deprecated version of the Office Fabric JS library. Couldn't find any documentation concerning the newest version (what they reference on the Office Fabric JS website) and my experiments weren't working. So I had to manually style it as best as I could. However, if you have an idea to solve this, would welcome it.

manifests/script-lab-edge.xml
Zlatkovsky: Being in app domains vs. being available on the UI are two completely different things.  App Domains just define the _whitelist_ of URLs that are going to be allowed to navigate to.   There is nothing wrong with defining more that we'll ever navigate to in practice.  So I think it's perfectly fine to add the localhost ones for all, just for purposes of keeping them as in-sync as possible for ease of diffing & copy-paste-ing.

src/client/app/components/about.ts
```diff
@@ -15,15 +15,15 @@ let { config } = PLAYGROUND;
                     <div class="profile__tertiary-text ms-font-m">User ID: ${storage.user}</div>
                     <div class="about__secondary-text ms-font-l">Version: {{config?.build?.version}}
                         <br/><span class="ms-font-m">(Deployed {{config?.build?.humanReadableTimestamp}})</span>
-                        <br/><span class="ms-font-m">{{config?.editorUrl}}</span>
                     </div>
-                    <div class="about__dropdown">
-                        <label class="ms-font-m about__dropdown-label">${Strings.aboutChangeEnvironment}</label>
-                        <select class="about__dropdown-select ms-font-m" [(ngModel)]="selectedConfig" (change)="changeConfig($event.target.value)">
+                    <pre class="about__tertiary-text ms-font-m">{{cache}}</pre>
```
Zlatkovsky: Not sure what it looks like without a screenshot showing it, but I'll trust your artistic judgement...

src/client/app/components/about.ts
```diff
@@ -41,45 +41,45 @@ export class About implements AfterViewInit {
     @Input() show: boolean;
     @Output() showChange = new EventEmitter<boolean>();
 
+    cache = `
```
Zlatkovsky: Using the template strings feels very weird here.  How about just putting these items as three array items, and doing a `.join('\n')`?

    cache = [
        Strings.aboutStorage,
        storageSize(localStorage, `playground_${environment.current.host}_snippets`, Strings.aboutSnippets),
        storageSize(sessionStorage, 'playground_intellisense', Strings.aboutIntellisense)
    ].join('\n');

src/client/app/components/about.ts
```diff
@@ -41,45 +41,45 @@ export class About implements AfterViewInit {
...
     ngAfterViewInit() {
         if (environment.current.config.name === config['local'].name) {
-            this.configs.push({ name: 'local', value: 'Localhost' });
+            this.configs.push({ name: 'local', value: config['local'].editorUrl });
         }
         this.configs.push(
             { name: 'production', value: 'Production' },
-            { name: 'edge', value: 'Beta' },
-            { name: 'insiders', value: 'Alpha' },
+            { name: 'edge', value: 'Alpha' },
```
Zlatkovsky: Could you order these in the order of **Production -> Beta -> Alpha** in the dropdown?  Also, move the "if" statement for localhost down after this block here, so that for localhost it goes **Production -> Beta -> Alpha -> https://localhost:3000**

src/client/app/components/about.ts
cshashwat: Screenshots have been updated in the pull request description. For easy reference, here's a picture of the about menu below:

capture3

config/env.config.js
```diff
@@ -55,5 +55,46 @@ const config = {
     }
 };
 
+const RedirectPlugin = function(options) { 
+    this.apply = function(compiler) {
+        compiler.plugin('compilation', function(compilation) {
+            compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
+                let htmlHead = htmlPluginData.html.match('<head>').index;
+                htmlPluginData.html = htmlPluginData.html.slice(0, htmlHead) + '<head>' +
```
dsaunders45459: Can you verify that <head> exists in the document before appending it. If for some reason the page does not have a <head> your script produces invalid HTML.

cshashwat:
Will add a ribbon in a later commit

WrathOfZombies:
@cshashwat this should help:https://gist.github.com/WrathOfZombies/2647965c07cec87b831d24fc75ab48df. Its written in ng1 but I am confident you'll be able to translate it to ng2 with ease.

manifests/script-lab-edge.xml
```diff
@@ -17,6 +17,8 @@
     <AppDomain>https://github.com</AppDomain>
     <AppDomain>https://stackoverflow.com</AppDomain>
     <AppDomain>https://dev.office.com</AppDomain>
+    <AppDomain>https://localhost:3000</AppDomain>
```
Zlatkovsky: Sorry to be nitpicky.  But just to maintain consistency and order everywhere, could all four files -- which now share the same app domains, right? -- have the app-domain ordering for the script lab domains be consistent in the right order?

I.e., PROD => BETA => ALPHA => LOCALHOST.

config/env.config.js
```diff
@@ -55,5 +55,46 @@ const config = {
...
+                            var queryExp = new RegExp("[\\?&]"+name+"=([^&#]*)", "i");
+                            var match = queryExp.exec(url);
+                            if (match && match.length > 1) {
+                                return match[1];
+                            }
+                            return "";
+                        };
+                        var originUrl = decodeURIComponent(getParameterByName("originEnvironment"));
+                        var newUrl = window.localStorage.getItem("environmentConfig");
+                        if (newUrl && !window.location.href.match(newUrl) && originUrl !== newUrl) {
```
dsaunders45459: I believe this is going to create a redirect loop where you have prod pointing to edge, edge pointing to insider, then try to navigate from insider back to prod. Please think through how to solve this issue or test to make sure we do not get in any loop.

manifests/script-lab-prod.xml
```diff
@@ -17,7 +17,14 @@
     <AppDomain>https://github.com</AppDomain>
     <AppDomain>https://stackoverflow.com</AppDomain>
     <AppDomain>https://dev.office.com</AppDomain>
-    <AppDomain>https://script-lab-runner.azureedge.net</AppDomain>  
+    <AppDomain>https://localhost:3000</AppDomain>
```
dsaunders45459: I am not sure we should be adding localhost to the production manifest. Office store team may have an issue with this in validation and it could be a security issue.

src/client/app/components/about.ts
```diff
@@ -12,10 +14,16 @@ import { environment, storageSize, Strings, storage } from '../helpers';
...
                     <div class="about__secondary-text ms-font-l">Version: {{config?.build?.version}}
-                    <br/><span class="ms-font-m">(Deployed {{config?.build?.humanReadableTimestamp}})</span>
-                    <br/><span class="ms-font-m">{{config?.editorUrl}}</span>
+                        <br/><span class="ms-font-m">(Deployed {{config?.build?.humanReadableTimestamp}})</span>
                     </div>
                     <pre class="about__tertiary-text ms-font-m">{{cache}}</pre>
+                    <div class="about__environment">
+                        <span class="ms-font-m about__environment-text">{{config?.editorUrl}}</span>
+                        <br/><label class="ms-font-m about__environment-text">${Strings.aboutSwitchEnvironment}</label>
+                        <select class="about__environment-select ms-font-m" [(ngModel)]="selectedConfig" (change)="changeConfig($event.target.value)">
```
dsaunders45459: I think we should be using ng-change instead of change

src/client/app/components/about.ts
```diff
@@ -29,17 +37,62 @@ import { environment, storageSize, Strings, storage } from '../helpers';
...
+            let result = await this._effects.alert(
+                Strings.changeEnvironmentConfirm,
+                `${Strings.aboutSwitchEnvironment} ${this.configs.find(c => c.name === currentConfigName).value } to ${this.configs.find(c => c.name === this.selectedConfig).value }`,
+                Strings.okButtonLabel,
+                Strings.cancelButtonLabel
+            );
+
+            if (result === Strings.cancelButtonLabel) {
+                this.selectedConfig = this.configs.find(c => c.name === currentConfigName).name;
+                this._changeDetector.detectChanges();
```
dsaunders45459: detectChanges may not be necessary if we use ng-change instead of change, but I am not an Angular expert

src/client/app/components/about.ts
```diff
@@ -29,17 +37,62 @@ import { environment, storageSize, Strings, storage } from '../helpers';
...
+        private _changeDetector: ChangeDetectorRef
+    ) { }
+
+    ngAfterViewInit() {
+        this.configs.push(
+            { name: 'production', value: 'Production' },
+            { name: 'insiders', value: 'Beta' },
+            { name: 'edge', value: 'Alpha' },
+        );
+        if (environment.current.config.name === config['local'].name) {
```
dsaunders45459: Now that we are passing originEnvironment it should be easy to add localhost to the list if user navigated from localhost to a different environment.

manifests/script-lab-prod.xml
Zlatkovsky: I think it makes it easier if we do have it everywhere (and esp. based on what you and I discussed, about being able to go back from production to localhost if you had originally sideloaded the dev one).  I don't see how it can be a security risk; the worst the user could do is hack their own selves.

But I'm also fine with removing it for prod.

config/env.config.js
dsaunders45459: Can you run the approach to solve this issue by @Zlatkovsky and I before starting on the implementation.

dsaunders45459:
@WrathOfZombies https://gist.github.com/WrathOfZombies/2647965c07cec87b831d24fc75ab48df is helpful, but I don't believe it solves the issue we are facing with Office UI Fabric 1.4 not letting us programmatically set the selected value of the dropdown (unless we reorder the current environment to first item). If this really is a bug, I think we should avoid the control as it also requires us to import an additional JavaScript library and impact perf.

config/webpack.common.js
```diff
@@ -155,6 +155,11 @@ module.exports = (prodMode) =>
                 filename: 'heartbeat.html',
                 template: './views/heartbeat.html',
                 chunks: ['polyfills', 'vendor', 'heartbeat'],
-            })
+            }),
+            new HtmlWebpackPlugin({
+                filename: 'tutorial.html',
+                template: './views/tutorial.html',
+            }),
+            new RedirectPlugin({})
```
Zlatkovsky: If you're adding the tutorial here, you should also remove it from above:

               {
                    from: './views/tutorial.html',
                    to: 'tutorial.html',
                },

config/env.config.js
```diff
@@ -59,8 +59,10 @@ const RedirectPlugin = function(options) {
...
         compiler.plugin('compilation', function(compilation) {
             compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
-                let htmlHead = htmlPluginData.html.match('<head>').index;
-                htmlPluginData.html = htmlPluginData.html.slice(0, htmlHead) + '<head>' +
+                let htmlHead = htmlPluginData.html.match('<head>');
+                if (htmlHead) {
+                    htmlHead = htmlHead.index;
+                    htmlPluginData.html = htmlPluginData.html.slice(0, htmlHead) + '<head>' +
                     `<script>
                         var getParameterByName = function(name) {
```
Zlatkovsky: Should this whole script be wrapped in a self-executing function, to avoid polluting the namespace?

     (function() {
            ...
     })();

Zlatkovsky: Is there a reason to declare it as a var, rather than `function getParameterByName() { ... }`

config/env.config.js
```diff
@@ -69,26 +71,38 @@ const RedirectPlugin = function(options) {
...
-                        var newUrl = window.localStorage.getItem("environmentConfig");
-                        if (newUrl && !window.location.href.match(newUrl) && originUrl !== newUrl) {
+                        var originUrl = getParameterByName("originEnvironment");
+                        var newUrl = getParameterByName("newEnvironment");
+                        if (newUrl) {
+                            newUrl = decodeURIComponent(newUrl);
+                            if (window.location.href.match(newUrl)) {
+                                window.localStorage.removeItem("environmentConfig");
+                            } else {
+                                window.localStorage.setItem("environmentConfig", newUrl);
```
Zlatkovsky: to follow the naming convention of the other local-storage items, can we call it `playground_environment_config`.  And could you make it a constant?  Also, I don't know if `config` is quite the right word (I imagine config to be a set of properties as a JSON object or something).  Maybe just a `playground_environment_redirect_url`, or similar?

config/env.config.js
```diff
@@ -69,26 +71,38 @@ const RedirectPlugin = function(options) {
...
                         };
-                        var originUrl = decodeURIComponent(getParameterByName("originEnvironment"));
-                        var newUrl = window.localStorage.getItem("environmentConfig");
-                        if (newUrl && !window.location.href.match(newUrl) && originUrl !== newUrl) {
+                        var originUrl = getParameterByName("originEnvironment");
+                        var newUrl = getParameterByName("newEnvironment");
+                        if (newUrl) {
+                            newUrl = decodeURIComponent(newUrl);
+                            if (window.location.href.match(newUrl)) {
+                                window.localStorage.removeItem("environmentConfig");
```
Zlatkovsky: Could you add a comment for what is happening here?  Why are you removing the setting (as opposed to, say, keeping it set to the new environment)?  There may well be a good reason for it, I just don't understand it...

Zlatkovsky: Oh, is it for the line `if (window.localStorage.getItem("environmentConfig")) {` below?  Would it make sense to just instead exit the function?  If you describe the flow in a comment (is it only the original manifest-pointed URL that needs to have the redirect config, or do they all?...) that would help.

config/env.config.js
```diff
@@ -55,5 +55,60 @@ const config = {
...
 
+const RedirectPlugin = function(options) { 
+    this.apply = function(compiler) {
+        compiler.plugin('compilation', function(compilation) {
+            compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
+                let htmlHead = htmlPluginData.html.match('<head>');
+                if (htmlHead) {
+                    htmlHead = htmlHead.index;
+                    htmlPluginData.html = htmlPluginData.html.slice(0, htmlHead) + '<head>' +
+                    `<script>
```
Zlatkovsky: Nit: I think you want the tick mark to be at the top of the previous line.  That way the produced html aligns nicely (or else the script tag has no tab, but everything else does.

config/env.config.js
```diff
@@ -55,5 +55,60 @@ const config = {
...
+                            if (window.location.href.match(newUrl)) {
+                                window.localStorage.removeItem("environmentConfig");
+                            } else {
+                                window.localStorage.setItem("environmentConfig", newUrl);
+                            }
+                        }
+
+                        if (window.localStorage.getItem("environmentConfig")) {
+                            var originParam = "" + 
+                            (window.location.search ? "&" : "?") +
```
Zlatkovsky: Indentation is odd here, as is below.  Also, what about using `[...].join()` instead?

src/client/app/components/about.ts
```diff
@@ -13,15 +15,21 @@ import {Strings, getAvailableLanguages, getDisplayLanguage, setDisplayLanguage }
...
                     </div>
                     <pre class="about__tertiary-text ms-font-m">{{cache}}</pre>
                     <div class="about__language">
                         <select class="about__language-select ms-font-m" [(ngModel)]="currentChosenLanguage" (change)="setLanguage($event.target.value)">
                             <option *ngFor="let l of availableLanguages" [value]="l.value">{{l.name}}</option>
                         </select>
                     </div>
+                    <div class="about__environment">
+                        <span class="ms-font-m about__environment-text">{{config?.editorUrl}}</span>
+                        <br/><label class="ms-font-m about__environment-text">${Strings().aboutSwitchEnvironment}</label>
```
Zlatkovsky: I don't think this (`${Strings().aboutSwitchEnvironment}`) will work.  In fact, I'm kind of suprised it didn't fail on you at runtime.  You should use it like this instead:  `{{strings.aboutSwitchEnvironment}}`

src/client/app/components/about.ts
Zlatkovsky: Ping.  Don't think you addressed this comment yet.

src/client/app/components/about.ts
```diff
@@ -50,10 +64,31 @@ export class About implements AfterViewInit {
...
         this.currentChosenLanguage = getDisplayLanguage();
         this.originalLanguage = this.currentChosenLanguage;
+        this.configs.push(
+            { name: 'production', value: 'Production' },
+            { name: 'insiders', value: 'Beta' },
+            { name: 'edge', value: 'Alpha' },
+        );
+
+        // user can only navigate to localhost if they've sideloaded local manifest
+        if (environment.current.config.name === config['local'].name ||
```
Zlatkovsky: Nit: can you isolate the large parenthetical inside the `if` statement into its own variable.  I.e., 

    showLocalConfig = ...;
    if (showLocalConfig) { ... }

src/client/app/components/about.ts
```diff
@@ -50,10 +64,31 @@ export class About implements AfterViewInit {
...
+        private _effects: UIEffects,
+        private _changeDetector: ChangeDetectorRef
+    ) {}
+
     ngAfterViewInit() {
         this.availableLanguages = getAvailableLanguages();
         this.currentChosenLanguage = getDisplayLanguage();
         this.originalLanguage = this.currentChosenLanguage;
+        this.configs.push(
+            { name: 'production', value: 'Production' },
```
Zlatkovsky: Nit: I wonder if you use the word "name" and "value" differently here than is customary.  Value is generally the programmatic name, whereas name is the display string.  I think you'd doing the opposite here.

E.g., instead of

    <option *ngFor="let c of configs" [value]="c.name">{{c.value}}</option>

It should really be:

    <option *ngFor="let c of configs" [value]="c.value">{{c.name}}</option>

src/client/app/components/about.ts
```diff
@@ -65,13 +100,34 @@ export class About implements AfterViewInit {
...
-    ${Strings().aboutStorage}
-    ${storageSize(localStorage, `playground_${environment.current.host}_snippets`, Strings().aboutSnippets)}
-    ${storageSize(sessionStorage, 'playground_intellisense', Strings().aboutIntellisense)}
-    `;
-
     setLanguage(languageCode: string) {
         this.currentChosenLanguage = languageCode;
     }
+
+    async changeConfig(newConfig: string) {
```
Zlatkovsky: Do you want to show the dialog only on "okClicked", like I did with the languages?  That way, if the user goes through the dropdown with a keyboard, they don't get inundated with alerts?

src/client/app/components/about.ts
```diff
@@ -65,13 +100,34 @@ export class About implements AfterViewInit {
...
+                this.strings.cancelButtonLabel
+            );
+
+            if (result === this.strings.cancelButtonLabel) {
+                this.selectedConfig = this.configs.find(c => c.name === currentConfigName).name;
+                this._changeDetector.detectChanges();
+                return;
+            }
+
+            let originEnvironment = window.localStorage.getItem('originEnvironment');
```
Zlatkovsky: And just to check, origin environment will always be a "clean" URL with no query parameters on it?  What if you were testing in the web, where you can have `https://localhost:3000/?mode=web`

Zlatkovsky: Basically, I wonder if you want to do a more robust query-parameter-merging, rather than assume that you can always stick a `?` at the end.

WrathOfZombies:
@dsaunders45459 I completely understand what you are getting at. Its actually not a bug but rather a feature limitation. You can programatically set the value but adding all the necessary classes and setting the value of the select box. Which is why a wrapper is required.  Apparently Fabric React solves this.

In any case, personally I don't think its a good idea to break UI patterns. If there's another way we can resolve this, we should. Its also why we only use Office Fabric core and write our own implementations.

Zlatkovsky:
@WrathOfZombies , how bad is it, though, UX-wise?  I think it looks quite in style with the other controls, there was nothing that immediately jumped out at me as being out of place.

If Fabric isn't actively working on addressing the feature limitation (and assuming they know about it; do they?), and if it looks OK as is, I would rather have @cshashwat work on some of the cool upcoming features.  it's not like many people will be going to the "about" dialog anyway...

WrathOfZombies:
@Zlatkovsky If @jakobpn is ok with that then go ahead 👍

config/env.config.js
cshashwat: Done. Thanks for pointing this out.

cshashwat: Fixed.

config/env.config.js
cshashwat: Added.

config/env.config.js
cshashwat: Added comments to clarify logic and  took your feedback into account from yesterday. Hopefully they make the general flow clear now.

config/env.config.js
cshashwat: Added constants for both `localStorage` parameters.

config/env.config.js
cshashwat: Added.

src/client/app/components/about.ts
cshashwat: This must have been overwritten when I merged the new changes, fixed now.

src/client/app/components/about.ts
cshashwat: Added.

src/client/app/components/about.ts
cshashwat: Fixed.

src/client/app/components/about.ts
cshashwat: Done.

src/client/app/components/about.ts
cshashwat: Added this functionality. User now selects new environment and then after "okClicked" the user gets the confirmation message that I had earlier. Clicking "Cancel" will reset the dropdown to current environment.

src/client/app/components/about.ts
cshashwat: Talked about this offline. Since we're using the set `editorUrl` from our defined `config` variable, adding a query string parameter in this manner should be fine.

config/env.config.js
```diff
@@ -73,34 +79,47 @@ const RedirectPlugin = function(options) {
...
 
-                        if (originUrl && !window.localStorage.getItem("originEnvironment")) {
-                            window.localStorage.setItem("originEnvironment", decodeURIComponent(originUrl));
+                        // Point app environment back to origin if user is not in origin
+                        if (originUrl.length > 0) {
+                            window.localStorage.setItem("${PLAYGROUND_ORIGIN}", decodeURIComponent(originUrl).toLowerCase());
                         }
-                    </script>` + 
+
+                        // Environment is already configured
```
Zlatkovsky: Nit, I would add a "**If reached here**, environment ...".  That way it's clear it's a complete sentence, and not a comment that got left accidentally from some refactoring.

src/client/app/strings/index.ts
```diff
@@ -165,6 +165,7 @@ export interface ClientStrings {
     aboutStorage: string;
     aboutSnippets: string;
     aboutIntellisense: string;
+    aboutCurrentEnvironment: string;
```
Zlatkovsky: Note, with the changes that I just merged from Mourad (German localization), your code will fail once you merge in his changes.  Follow the pattern that I did for `snippetInfoDialogTitle: englishSubstitutesForNotYetTranslated.snippetInfoDialogTitle,` in German.ts in the client strings.

config/env.config.js
```diff
@@ -55,5 +55,81 @@ const config = {
...
+const PLAYGROUND_ORIGIN = "playground_origin_environment_url";
+
+const PLAYGROUND_REDIRECT = "playground_redirect_environment_url";
+
+const RedirectPlugin = function(options) { 
+    this.apply = function(compiler) {
+        compiler.plugin('compilation', function(compilation) {
+            compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
+                let htmlHead = htmlPluginData.html.match('<head>');
+                if (htmlHead) {
```
dsaunders45459: Lets store <head> in a const

config/env.config.js
```diff
@@ -55,5 +55,81 @@ const config = {
...
+                            if (window.location.href.toLowerCase().indexOf(targetUrl) != -1) {
+                                window.localStorage.removeItem("${PLAYGROUND_REDIRECT}");
+                                return;
+                            }
+
+                            window.localStorage.setItem("${PLAYGROUND_REDIRECT}", targetUrl);
+                        }
+
+                        // Redirect origin environment to target
+                        if (window.localStorage.getItem("${PLAYGROUND_REDIRECT}")) {
```
dsaunders45459: We are making multiple calls to getItem with the same key. Better performance to store it in a local variable

config/env.config.js
dsaunders45459: It is used 3 times

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+    constructor(
+        private _effects: UIEffects,
+        private _changeDetector: ChangeDetectorRef
+    ) {}
+
     ngAfterViewInit() {
         this.availableLanguages = getAvailableLanguages();
         this.currentChosenLanguage = getDisplayLanguage();
         this.originalLanguage = this.currentChosenLanguage;
+        this.configs.push(
```
dsaunders45459: It would be better to reduce coupling and populate this by iterating through environment configs.

config/env.config.js
cshashwat: Added.

config/env.config.js
cshashwat: Added.

src/client/app/components/about.ts
cshashwat: Discussed offline. Since we want the custom order, will keep as is for now.

src/client/app/components/about.ts
```diff
@@ -13,15 +16,20 @@ import {Strings, getAvailableLanguages, getDisplayLanguage, setDisplayLanguage }
...
                     <pre class="about__tertiary-text ms-font-m">{{cache}}</pre>
                     <div class="about__language">
                         <select class="about__language-select ms-font-m" [(ngModel)]="currentChosenLanguage" (change)="setLanguage($event.target.value)">
                             <option *ngFor="let l of availableLanguages" [value]="l.value">{{l.name}}</option>
                         </select>
                     </div>
+                    <div class="about__environment">
+                        <label class="ms-font-m about__environment-text">{{strings.aboutCurrentEnvironment}}</label>
+                        <select class="about__environment-select ms-font-m" [(ngModel)]="selectedConfig" (ngModelChange)="changeConfig($event)">
+                            <option *ngFor="let c of configs" [value]="c.value">{{c.name}}</option>
```
WrathOfZombies: NIT: variable naming, elaborate `c`

src/client/app/components/about.ts
WrathOfZombies: Why not just initialize the array at L66?

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
         this.currentChosenLanguage = getDisplayLanguage();
         this.originalLanguage = this.currentChosenLanguage;
+        this.configs.push(
+            { name: 'Production', value: 'production' },
+            { name: 'Beta', value: 'insiders' },
+            { name: 'Alpha', value: 'edge' },
+        );
+
+        // User can only navigate to localhost if they've sideloaded local manifest
+        let showLocalConfig = environment.current.config.name === config['local'].name || (window.localStorage.getItem(PLAYGROUND_ORIGIN) && window.localStorage.getItem(PLAYGROUND_ORIGIN).indexOf('localhost') !== -1);
```
WrathOfZombies: use the storage service or settings and I'd rather do `/localhost/i.test` instead of `indexOf`

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
-    ${storageSize(sessionStorage, 'playground_intellisense', Strings().aboutIntellisense)}
-    `;
+        let currentConfigName = environment.current.config.name.toLowerCase();
+        if (this.selectedConfig === currentConfigName) {
+            return;
+        }
+
+        let result = await this._effects.alert(
+            this.strings.changeEnvironmentConfirm,
+            `${this.strings.aboutSwitchEnvironment} ${this.configs.find(c => c.value === currentConfigName).name } to ${this.configs.find(c => c.value === this.selectedConfig).name }`,
```
WrathOfZombies: NIT: Variable naming

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+            this.strings.okButtonLabel,
+            this.strings.cancelButtonLabel
+        );
+        if (result === this.strings.cancelButtonLabel) {
+            this.selectedConfig = this.configs.find(c => c.value === currentConfigName).value;
```
WrathOfZombies: NIT: Variable naming

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+            this._changeDetector.detectChanges();
+            return;
+        }
+
+        let originEnvironment = window.localStorage.getItem(PLAYGROUND_ORIGIN);
+        let targetEnvironment = config[this.selectedConfig].editorUrl;
+
+        // Add query string parameters to default editor URL
+        if (originEnvironment) {
+            window.location.href = originEnvironment + '?targetEnvironment=' + encodeURIComponent(targetEnvironment);
```
WrathOfZombies: NIT: Use template strings instead

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+        } else {
+            window.localStorage.setItem(PLAYGROUND_REDIRECT, targetEnvironment);
+            window.location.href = targetEnvironment + '?originEnvironment=' + encodeURIComponent(environment.current.config.editorUrl);
```
WrathOfZombies: NIT: Use template strings instead

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+        }
+    }
 
     setLanguage(languageCode: string) {
         this.currentChosenLanguage = languageCode;
     }
+
+    changeConfig(newConfig: string) {
```
WrathOfZombies: Do you need this function?

src/client/app/components/about.ts
```diff
@@ -50,28 +63,75 @@ export class About implements AfterViewInit {
...
+        // Add query string parameters to default editor URL
+        if (originEnvironment) {
+            window.location.href = originEnvironment + '?targetEnvironment=' + encodeURIComponent(targetEnvironment);
+        } else {
+            window.localStorage.setItem(PLAYGROUND_REDIRECT, targetEnvironment);
+            window.location.href = targetEnvironment + '?originEnvironment=' + encodeURIComponent(environment.current.config.editorUrl);
+        }
+    }
 
     setLanguage(languageCode: string) {
```
WrathOfZombies: Do you need this function instead of an inline expression?

Zlatkovsky: Yeah, that was my bad (I had originally had more logic there, but then moved it out and left as a one-liner.  @cshashwat , please go ahead and make the change both for `setLanguage` and `changeConfig`

src/client/app/components/about.ts
Zlatkovsky: Nit: By the way, you should also break apart this line into multiple lines for readability.

src/client/app/components/about.ts
cshashwat: These functions are actually not needed. The ngModel binding in the `select` tag automatically updates `selectedConfig` (and `currentChosenLanguage` in the above `select`). Since the 'okClicked()' function is actually handling any events that we need, both functions are superfluous.

src/client/app/components/about.ts
cshashwat: See above comment.

src/client/app/components/about.ts
cshashwat: Not sure exactly what you mean by this. Do you mean when I'm building the new URL? How would a template string differ from the current concatenation?

src/client/app/components/about.ts
cshashwat: Fixed.

src/client/app/components/about.ts
cshashwat: Fixed.

src/client/app/components/about.ts
cshashwat: I actually removed one of the conditions, so one line should be enough now? Can fix if still necessary.

src/client/app/components/about.ts
cshashwat: Fixed.

src/client/app/components/about.ts
```diff
@@ -1,6 +1,10 @@
-import { Component, Input, ChangeDetectionStrategy, Output, EventEmitter, AfterViewInit } from '@angular/core';
+import { Component, Input, ChangeDetectionStrategy, Output, EventEmitter, AfterViewInit, ChangeDetectorRef } from '@angular/core';
 import { environment, storageSize, storage } from '../helpers';
 import {Strings, getAvailableLanguages, getDisplayLanguage, setDisplayLanguage } from '../strings';
+import { UIEffects } from '../effects/ui';
+const PLAYGROUND_ORIGIN = 'playground_origin_environment_url';
```
cshashwat: There is a duplicate here in constants between this file and the `env.config.js` file. For some reason, the build was failing when I tried to import the constants in `about.ts` via the line

`const { PLAYGROUND_ORIGIN, PLAYGROUND_REDIRECT } = require('../../../../config/env.config');`

The pathname was correct, but it would fail. If there's a better way to do it, would welcome thoughts.

src/client/app/components/about.ts
Zlatkovsky: Bhargav means: "${originEnvironment}?${targetEnvironment}=${encodeURIComponent(targetEnvironment)}"

cshashwat: Added.

src/client/app/components/about.ts
WrathOfZombies: I am not seeing it...
image

WrathOfZombies: Nevermind.. GitHub works differently as opposed to VSO.

src/client/app/components/about.ts
WrathOfZombies: Which is why we have an `environment.ts`. The idea is that we use webpack to inject a constant based on the env that it was built for. So you cant access `config/env.config` cause that's outside the projects build folder.

You can access the constants from https://github.com/OfficeDev/script-lab/blob/master/src/client/app/helpers/environment.ts where `PLAYGROUND` is injected in via https://github.com/OfficeDev/script-lab/blob/master/config/webpack.common.js#L69 which gets the build info and config from https://github.com/OfficeDev/script-lab/blob/master/config/webpack.common.js#L9

src/client/assets/styles/components/_about.scss
```diff
@@ -32,8 +32,37 @@
...
+        margin-top: 5px;
+        padding: 5px;
+        width: 60%;
+        outline: 0;
+        border: 1px solid grey;
+
+        /* arrow styling */
+        -webkit-appearance: none;
+        appearance: none;
+        background: linear-gradient(grey, grey) no-repeat, linear-gradient(-135deg, rgba(255, 255, 255, 0) 50%, white 50%) no-repeat, linear-gradient(-225deg, rgba(255, 255, 255, 0) 50%, white 50%) no-repeat, linear-gradient(black,grey) no-repeat;
```
WrathOfZombies: You can collapse the `background` into a short hand.

src/client/public/index.script.ts
```diff
@@ -11,6 +13,31 @@ import { Strings } from '../app/strings';
     document.getElementById('choose-your-host').textContent = strings.HtmlPageStrings.chooseYourHost;
     document.getElementById('choose-your-host').style.visibility = 'visible';
 
+    let ribbons = document.getElementsByClassName('ribbon');
+    for (let i = 0; i < ribbons.length; i++) {
```
WrathOfZombies: Dont do `ribbons.length`. Access the `length` property is more complex. Given modern machines you'd still be able to getaway using `ribbons.forEach(ribbon => {})`. If you absolutely need max perf, then go ahead and do:

`for (const ribbon of ribbons) { }`

src/client/public/run.script.ts
```diff
@@ -0,0 +1,29 @@
+import { environment } from '../app/helpers';
+let { config } = PLAYGROUND;
+
+(() => {
+    let ribbons = document.getElementsByClassName('ribbon');
+    for (let i = 0; i < ribbons.length; i++) {
```
WrathOfZombies: modify the for-loop as mentioned above.

src/client/views/index.html
```diff
@@ -35,6 +35,8 @@
 	<script src="https://appsforoffice.microsoft.com/lib/1.1/hosted/office.js"></script>
 	<app class="app">
 		<div id="progress" class="ms-progress-component">
+			<div class="ribbon left"></div>
```
WrathOfZombies: Do we really need am empty `div`. If that's the case I feel you should be able to just do `#progress::before` & `#progress::after` instead

src/client/public/index.script.ts
```diff
@@ -11,6 +13,31 @@ import { Strings } from '../app/strings';
     document.getElementById('choose-your-host').textContent = strings.HtmlPageStrings.chooseYourHost;
```
WrathOfZombies: cache this into a variable and reuse

config/env.config.js
```diff
@@ -55,5 +57,77 @@ const config = {
     }
 };
 
+const RedirectPlugin = function(options) {
```
WrathOfZombies: Use lambda's instead unless you absolutely need the `this` to reference the `window/executing function`

config/env.config.js
```diff
@@ -55,5 +57,77 @@ const config = {
...
+                            window.localStorage.setItem("${PLAYGROUND_ORIGIN}", decodeURIComponent(originUrl).toLowerCase());
+                        }
+
+                        // If reached here, environment is already configured
+                    })();
+                    </script>
+                    ` + 
+                    htmlPluginData.html.slice(htmlHead + headOpeningTag.length);
+                }
+                callback(null, htmlPluginData);
```
WrathOfZombies: Can we do a `promise` instead?

config/env.config.js
```diff
@@ -55,5 +57,77 @@ const config = {
     }
 };
 
+const RedirectPlugin = function(options) { 
+    this.apply = function(compiler) {
+        compiler.plugin('compilation', function(compilation) {
+            compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPluginData, callback) {
+                let headOpeningTag = '<head>'; 
+                let htmlHead = htmlPluginData.html.match(headOpeningTag);
+                if (htmlHead) {
```
WrathOfZombies: Ensure htmlHead is always > 0 else `slice` can have interesting effects

config/env.config.js
```diff
@@ -55,5 +57,77 @@ const config = {
...
+                    <script>
+                    (function() {
+                        function getParameterByName(name) {
+                            var url = window.location.search;
+                            var queryExp = new RegExp("[\\?&]"+name+"=([^&#]*)", "i");
+                            var match = queryExp.exec(url);
+                            if (match && match.length > 1) {
+                                return match[1];
+                            }
+                            return null;
```
WrathOfZombies: return '' instead else your toLowerCase() will fail.

config/env.config.js
```diff
@@ -55,5 +57,77 @@ const config = {
...
+                        // Redirect origin environment to target
+                        var redirectUrl = window.localStorage.getItem("${PLAYGROUND_REDIRECT}");
+                        if (redirectUrl) {
+                            var originParam = [
+                                (window.location.search ? "&" : "?"), 
+                                "originEnvironment=",
+                                encodeURIComponent(window.location.origin)
+                            ].join("");
+
+                            window.location.href = [
```
WrathOfZombies: use `location.replace` instead of updating href property manually

WrathOfZombies:
@Zlatkovsky there's something wrong with our bundle output. 
image

`error, tutorialScript, runScript, indexScript, heartbeat` have a lot of code duplication given that their sizes are almost the same consistently and the source code isn't huge.

Zlatkovsky:
@WrathOfZombies , if you can help @cshashwat with figuring out what's going with bunding, that would be great!  I merely followed the pattern of what we already had for functions and heartbeat, which I assume you had put in there months ago.

config/env.config.js
cshashwat: This should not be an issue, as it is checked in the function a few lines down, via the statement

`
(getParameterByName("originEnvironment") || "").toLowerCase();
`

Wanted the internal function to return null to explicitly signal that the parameter did not exist.

src/client/views/index.html
cshashwat: I believe this is necessary to populate the div with the environment contents on page load. In `index.script.tx` and `run.script.ts`, these divs are filled via `textContent`, and the user will see the left and right ribbons with the appropriate environment when loading the app. Not sure how I would accomplish the same with `before` and `after` (also have not used them before).

src/client/public/index.script.ts
cshashwat: Tried this already: due to the nature of `HTMLCollection` , `forEach` or normal array iteration won't work. The only property exposed seems to be `length`, which is why I'm using it:

https://stackoverflow.com/questions/22754315/for-loop-for-htmlcollection-elements

config/env.config.js
cshashwat: Added.

src/client/public/index.script.ts
cshashwat: Added.

src/client/app/components/about.ts
cshashwat: Thanks for pointing this out. Added.

config/env.config.js
cshashwat: Added.

config/env.config.js
cshashwat: Added.

config/env.config.js
cshashwat: Do you mean for the `callback()` function? The code I used to create my personal `RedirectPlugin` is derived from the HTMLPLugin documentation:

https://github.com/jantimon/html-webpack-plugin

config/env.config.js
cshashwat: Actually, since this is being initialized as a plugin in the `webpack.common` file, I don't believe I can use lambda notation? That would not allow me to have a constructor.

src/client/assets/styles/components/_about.scss
cshashwat: How would I collapse it further? I have multiple, different gradients, and I thought I already had them in short-hand notation. Would welcome any suggestions you have.

config/env.config.js
WrathOfZombies: Umm I should have worded it better. Declare RedirectPlugin as a class instead and use the method notation for `apply`. 
Inside of `apply` for the compiler.plugin use lambdas throughout.

config/env.config.js
WrathOfZombies: Ok missed that 👍

config/env.config.js
cshashwat: Added this now; should be using what you outlined.

config/env.config.js
cshashwat: The callback function is injected via the plugin, here in line:

`                
compilation.plugin('html-webpack-plugin-before-html-processing', (htmlPluginData, callback) =>
`

So I don't think we can use a promise.

Issue #458, opened 2017-06-26
Cannot "Export for publishing" on macOS


Hongbo-Miao:
## Current Behavior
When I click "Export for publishing",
![image](https://user-images.githubusercontent.com/3375461/27552462-5f4d6526-5a5c-11e7-99d7-cd8406b36231.png)

it shows this
![image](https://user-images.githubusercontent.com/3375461/27552470-68a16c44-5a5c-11e7-8893-286a18107cf7.png)

## Your Environment
* Operating System and version: macOS 10.12.5

Zlatkovsky:
If you were to do the same thing from the browser (https://script-lab.azureedge.net), does the issue occur there as well?  Or is it only because it's an add-in.

We use `FileSaver.js` for exporting.  Could you also see if these demos work for you:  https://eligrey.com/demos/FileSaver.js/?

Thanks!

Zlatkovsky:
@Hongbo-Miao , please take a look at the questions above.  thanks!

Hongbo-Miao:
Just tried, both https://script-lab.azureedge.net and https://eligrey.com/demos/FileSaver.js succeed download the file.

Zlatkovsky:
In that case, it might be a platform bug with agaves on Mac.  Could you please file a bug and follow up with Dan or Julian?  Thanks!

Hongbo-Miao:
@dsaunders45459 saw u self-assigned, let me know if still need notify Dan. thanks

dsaunders45459:
I thought this was an issue with Mac platform so I assigned to myself, but it seems to also repro on Safari. @Hongbo-Miao you can fix if you want if not a platform issue.

Hongbo-Miao:
@dsaunders45459 actually I am not sure what cause this since it works on my Safari. Please keep diging into it.

dsaunders45459:
I spent more time investigating this. Safari had the same bug before 10.1, the FileSaver library would not work for it. Mac Office uses the WebKit control to embed Script Lab, and it looks like it is not currently possible for the blob download approach FileSaver uses to work with this control using public WebKit API. An alternate approach could be to use window.open on a data-uri and base64 string for the contents, but FileSaver library does not do this. I have opened a bug in WebKit https://bugs.webkit.org/show_bug.cgi?id=174076 to enable blob downloads, but the fix will not be available in the foreseeable future. For now I think we should hide the button on the Mac platform, I will send out a pull request.

[...]

PR #467, opened 2017-07-03
Hide export for publishing button when running on Mac

dsaunders45459:
## Description
The download blob feature that FileSaver uses to save the zip is not working in WebKit. You can see more technical details in https://github.com/OfficeDev/script-lab/issues/458 and https://bugs.webkit.org/show_bug.cgi?id=174076. Since we are not able to fix this issue until the change has been implemented and shipped in WebKit, it is better to hide the "Export for publishing" button on the Mac platform instead of it always being a no-op.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/458

## Motivation and Context

If the user clicks the "Export for publishing" button currently, they think Script Lab is broken because it does not do anything. If the button does not show up the user will know the download feature is not yet supported on this platform and instead use the features that work.

## How Has This Been Tested?

I have verified that the "Export for publishing" shows up when loading the editor in the browser, but when I side load the manifest on Mac it is hidden.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.


src/client/app/containers/app.ts
```diff
@@ -23,7 +24,7 @@ import { isNil } from 'lodash';
                     <command icon="PageCheckedin" title="${Strings.shareMenuPublic}" (click)="shareGist({isPublic: true, isUpdate: false})"></command>
                     <command icon="ProtectedDocument" title="${Strings.shareMenuPrivate}" (click)="shareGist({isPublic: false, isUpdate: false})"></command>
                     <command id="CopyToClipboard" icon="Copy" title="${Strings.shareMenuClipboard}" (click)="shareCopy()"></command>
-                    <command icon="Download" title="${Strings.shareMenuExport}" (click)="shareExport()"></command>
+                    <command [hidden]="!isDownloadZipSupported" icon="Download" title="${Strings.shareMenuExport}" (click)="shareExport()"></command>
```
Zlatkovsky: I'm not in love with hiding the button.  The developer has no idea why it's not available, or they might watch a video and see it working for someone, but then not see it themselves.

I would prefer to still keep the button, but then fail with an error message saying that export on Mac unfortunately isn't supported; but that the user can export from a non-Mac machine if they wish.

src/client/app/containers/app.ts
```diff
@@ -78,6 +79,10 @@ export class AppComponent {
             );
     }
 
+    get isDownloadZipSupported() {
+        return Utilities.platform !== PlatformType.MAC;
```
Zlatkovsky: I'm assuming Mac & iPad are both impacted, right?
Admittedly, we aren't really testing iPad.  But the check may as well be on both, I would think...

src/client/app/containers/app.ts
WrathOfZombies: Agreed, makes sense. Or you can grey it out with a title saying this feature is not supported.

src/client/app/effects/github.ts
```diff
@@ -9,9 +9,10 @@ import { Effect, Actions } from '@ngrx/effects';
 import { Http, ResponseContentType } from '@angular/http';
 import * as clipboard from 'clipboard';
 import { UIEffects } from './ui';
-import { find } from 'lodash';
+import { find, isNil } from 'lodash';
```
WrathOfZombies: Do you need `isNil`? you can just do `(exception == null)` that checks for both null and undefined [literally the implementation of isNil].

Zlatkovsky:
@dsaunders45459 , ping.

src/client/app/effects/github.ts
Zlatkovsky: I personally like the isNil simply for its explicitness (I always worry about seeing a double-equals, and feeling like maybe it will trigger for other falsy things, not just null and undefined)

PR #473, opened 2017-07-06
Add to update feature: existing snippets with gist ids are assumed to be updateable until update request fails

cshashwat:
## Description

Existing snippets in users' local storage, if they were imported from GitHub, have gist IDs associated with them, but they do not have the new `gistOwnerId` field added in the update feature. As a result users will currently not see the 'Update' option.

This pull request optimistically assumes that previously imported gists are owned by the user, so the user will now see the 'Update' option. If the user tries to update the gist and fails, then the gist may not exist anymore or the user may not own it: the program looks for this error code and then sets `gistOwnerId` so that the user no longer sees the 'Update' option.

## Motivation and Context

Previously imported gists should be able to be updated, but right now that's not possible.

## How Has This Been Tested?

Has been tested via the different 'Update' scenarios in `TESTING.md`.

## Types of changes

## Checklist:

### DIFF
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -73,12 +73,11 @@ export class AppComponent {
         return this.profile$
             .filter(profile => profile != null)
             .map(profile => {
-                if (!isNil(this.snippet.gistOwnerId)) {
-                    return this.snippet.gistOwnerId === profile.login;
+                if (!isNil(this.snippet.gist)) {
+                    return !isNil(this.snippet.gistOwnerId) ? this.snippet.gistOwnerId === profile.login : true;
                 }
                 return false;
-            }
-            );
+            });
     }
 
     menuOpened$ = this._store.select(fromRoot.getMenu);
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -5,7 +5,7 @@ import { UI, Snippet, GitHub } from '../actions';
 import { UIEffects } from '../effects/ui';
 import { environment, isOfficeHost, isInsideOfficeApp } from '../helpers';
 import { Strings } from '../strings';
-import { isNil } from 'lodash';
+import { isEmpty } from 'lodash';
 
 @Component({
     selector: 'app',
@@ -73,12 +73,12 @@ export class AppComponent {
         return this.profile$
             .filter(profile => profile != null)
             .map(profile => {
-                if (isNil(this.snippet.gist)) {
+                if (isEmpty(this.snippet.gist)) {
                     return false;
                 }
 
                 // Assume that user owns gist, for back-compat
-                return isNil(this.snippet.gistOwnerId) ? true : this.snippet.gistOwnerId === profile.login;
+                return isEmpty(this.snippet.gistOwnerId) ? true : this.snippet.gistOwnerId === profile.login;
             });
     }
 
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -73,10 +73,12 @@ export class AppComponent {
         return this.profile$
             .filter(profile => profile != null)
             .map(profile => {
-                if (!isNil(this.snippet.gist)) {
-                    return !isNil(this.snippet.gistOwnerId) ? this.snippet.gistOwnerId === profile.login : true;
+                if (isNil(this.snippet.gist)) {
+                    return false;
                 }
-                return false;
+
+                // Assume that user owns gist, for back-compat
+                return isNil(this.snippet.gistOwnerId) ? true : this.snippet.gistOwnerId === profile.login;
             });
     }
 
--- a/src/client/app/effects/github.ts
+++ b/src/client/app/effects/github.ts
@@ -14,7 +14,7 @@ import * as moment from 'moment';
 import * as fromRoot from '../reducers';
 
 const FileSaver = require('file-saver');
-const UNKNOWN_GIST_OWNER_ID = '<unknown'; // GitHub IDs can only have alphanumeric characters
+const UNKNOWN_GIST_OWNER_ID = '<unknown>'; // Intentional use of brackets tnat are not allowed in a GitHub ID (can only have alphanumeric characters)
 
 @Injectable()
 export class GitHubEffects {
@@ -92,7 +92,7 @@ export class GitHubEffects {
             GitHub.GitHubActionTypes.UPDATE_GIST
         )
         .mergeMap(({ payload, type }) => {
-            let { id, name, description, gist } = payload;
+            let { id, name, description, gist, gistOwnerId } = payload;
             let files: IGistFiles = {};
             let gistId = null;
 
@@ -110,14 +110,14 @@ export class GitHubEffects {
             }
 
             return this._github.createOrUpdateGist(
-                    description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
+                description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
             )
             .map((gist: IGist) => ({ gist: gist, snippetId: id }))
             .catch(exception => {
-                if (exception.status === 404) {
-                    throw { id: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID};
+                if (!gistOwnerId && exception.status >= 400 && exception.status <= 499) {
+                    throw new Error(JSON.stringify({ id: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
                 }
-                throw null;
+                throw exception;
             });
         })
         .mergeMap(async ({ gist, snippetId }) => {
@@ -141,10 +141,14 @@ export class GitHubEffects {
                 new Snippet.UpdateInfoAction({ id: snippetId, gist: gist.id, gistOwnerId: gist.owner.login })])
         )
         .catch(exception => {
-            let { gistOwnerId, id } = exception;
-            let action = (gistOwnerId && id)
-            ? new Snippet.UpdateInfoAction({id: id, gistOwnerId: gistOwnerId})
-            : new GitHub.ShareFailedAction(exception);
+            if (exception instanceof Error) {
+                exception = JSON.parse(exception.message);
+            }
+            let { id, gistOwnerId } = exception;
+            let action =
+                (gistOwnerId && id) ?
+                    new Snippet.UpdateInfoAction({id: id, gistOwnerId: gistOwnerId}) :
+                    new GitHub.ShareFailedAction(exception);
 
             this._uiEffects.alert(
                 Strings().gistShareFailedBody + '\n\n' + Strings().reloadPrompt,
--- a/src/client/app/effects/github.ts
+++ b/src/client/app/effects/github.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
-import { AI, getShareableYaml, environment } from '../helpers';
+import { PlaygroundError, AI, getShareableYaml, environment } from '../helpers';
 import { Strings, getDisplayLanguage } from '../strings';
 import { GitHubService } from '../services';
 import { Store, Action } from '@ngrx/store';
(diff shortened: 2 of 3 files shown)


src/client/app/containers/app.ts
```diff
@@ -73,12 +73,11 @@ export class AppComponent {
         return this.profile$
             .filter(profile => profile != null)
             .map(profile => {
-                if (!isNil(this.snippet.gistOwnerId)) {
-                    return this.snippet.gistOwnerId === profile.login;
+                if (!isNil(this.snippet.gist)) {
+                    return !isNil(this.snippet.gistOwnerId) ? this.snippet.gistOwnerId === profile.login : true;
```
Zlatkovsky: could you reverse the logic so as not to have a "not" for the condition?

Zlatkovsky: And also a comment about the "true", saying that we'll optimistically assume "true" for any snippet that doesn't specify an owner, for back-compat / lightup.

src/client/app/effects/github.ts
```diff
@@ -14,6 +14,7 @@ import * as moment from 'moment';
 import * as fromRoot from '../reducers';
 
 const FileSaver = require('file-saver');
+const UNKNOWN_GIST_OWNER_ID = '<unknown'; // GitHub IDs can only have alphanumeric characters
```
Zlatkovsky: I assume you want a closing ">" as well?  Also, not sure what you mean by your comment.  Is it that you're including the brackets so as to make 100% sure that no user will ever have this name?  If so, makes sense, but maybe change the comment to "intentionally using brackets that are not allowed in a GitHub ID (which only allows alphanumeric characters)"

src/client/app/effects/github.ts
```diff
@@ -109,9 +110,15 @@ export class GitHubEffects {
             }
 
             return this._github.createOrUpdateGist(
-                description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
```
Zlatkovsky: Why the extra tab?

src/client/app/effects/github.ts
```diff
@@ -134,12 +141,17 @@ export class GitHubEffects {
                 new Snippet.UpdateInfoAction({ id: snippetId, gist: gist.id, gistOwnerId: gist.owner.login })])
         )
         .catch(exception => {
+            let { gistOwnerId, id } = exception;
+            let action = (gistOwnerId && id)
+            ? new Snippet.UpdateInfoAction({id: id, gistOwnerId: gistOwnerId})
```
Zlatkovsky: Indent these?  And I think the usual convention for this is for the "?" to be at the beginning of the previous line, as is the case with ":"

src/client/app/effects/github.ts
```diff
@@ -109,9 +110,15 @@ export class GitHubEffects {
...
 
             return this._github.createOrUpdateGist(
-                description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
+                    description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
             )
-            .map((gist: IGist) => ({ gist: gist, snippetId: id }));
+            .map((gist: IGist) => ({ gist: gist, snippetId: id }))
+            .catch(exception => {
+                if (exception.status === 404) {
+                    throw { id: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID};
```
Zlatkovsky: You should always throw instances of an Error object (JS best practice).  If this is an intentional error, which this one is, that's OK, just use or create your own intentional/informational error class.  See "class InformationalError extends Error {" in the server code for an example.

src/client/app/effects/github.ts
```diff
@@ -134,12 +141,17 @@ export class GitHubEffects {
                 new Snippet.UpdateInfoAction({ id: snippetId, gist: gist.id, gistOwnerId: gist.owner.login })])
         )
         .catch(exception => {
+            let { gistOwnerId, id } = exception;
```
Zlatkovsky: Will there ever be another exception that could get thrown here, that isn't your exception?  Should you do some sort of an instanceof check?  (Yes, you should; this ties in to the other comment above)

src/client/app/effects/github.ts
```diff
@@ -109,9 +110,15 @@ export class GitHubEffects {
             }
 
             return this._github.createOrUpdateGist(
-                description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
+                    description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
             )
-            .map((gist: IGist) => ({ gist: gist, snippetId: id }));
+            .map((gist: IGist) => ({ gist: gist, snippetId: id }))
+            .catch(exception => {
+                if (exception.status === 404) {
```
Zlatkovsky: Is it 404 according their documentation for unauthorized?  Do they use it for anything else like a missing ID?

Zlatkovsky: By the way, if the error doesn't provide you enough information, you may consider doing a pre-validation as a separate call (instead of watching failure on update, to first find the owner ID via an API call to github, and so handle it as normal logic rather than an exception).  I'm fine with either way, as long as we can have enough info.

src/client/app/containers/app.ts

src/client/app/effects/github.ts
cshashwat: Modified and added comment.

src/client/app/effects/github.ts
cshashwat: Didn't realize I had added that. Thanks for pointing it out, fixed.

src/client/app/effects/github.ts
cshashwat: Discussed offline. Keeping `<unknown>` for now, since it doesn't matter for the purposes of the app whether we populate it with this or with the other person's actual login, as the 'Update' won't show either way. Have modified logic to handle more error codes, in case GitHub changes them to be accurate, and have also made it so that if `gistOwnerId` is already set, we don't change it to `<unknown>`.

src/client/app/effects/github.ts
cshashwat: I added it as a new `Error`, with its message field containing the JSON object with appropriate parameters. The `exception` in the `catch` blocks in this function is normally of type `Response` so throwing something of type `Error` should be enough to distinguish it I believe.

src/client/app/effects/github.ts
cshashwat: Yup, added this.

src/client/app/effects/github.ts
cshashwat: Modified this to look like the multi-line statement in `scrubbedIfNeeded()` in `effects/snippet.ts`, but not sure if it's what you meant. Can change if necessary.

src/client/app/effects/github.ts
```diff
@@ -111,7 +112,13 @@ export class GitHubEffects {
             return this._github.createOrUpdateGist(
                 description, files, gistId, type === GitHub.GitHubActionTypes.SHARE_PUBLIC_GIST
             )
-            .map((gist: IGist) => ({ gist: gist, snippetId: id }));
+            .map((gist: IGist) => ({ gist: gist, snippetId: id }))
+            .catch(exception => {
+                if (!gistOwnerId && exception.status >= 400 && exception.status <= 499) {
+                    throw new Error(JSON.stringify({ id: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
```
dsaunders45459: I believe if key and value are the same in the object you can just use { id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }

src/client/app/effects/github.ts
```diff
@@ -14,6 +14,7 @@ import * as moment from 'moment';
 import * as fromRoot from '../reducers';
 
 const FileSaver = require('file-saver');
+const UNKNOWN_GIST_OWNER_ID = '<unknown>'; // Intentional use of brackets tnat are not allowed in a GitHub ID (can only have alphanumeric characters)
```
dsaunders45459: *that* spelling

src/client/app/effects/github.ts
```diff
@@ -134,12 +141,21 @@ export class GitHubEffects {
                 new Snippet.UpdateInfoAction({ id: snippetId, gist: gist.id, gistOwnerId: gist.owner.login })])
         )
         .catch(exception => {
+            if (exception instanceof Error) {
+                exception = JSON.parse(exception.message);
```
dsaunders45459: It is not good to change the type of exception from Error to object.

src/client/app/containers/app.ts
```diff
@@ -73,12 +73,13 @@ export class AppComponent {
         return this.profile$
             .filter(profile => profile != null)
             .map(profile => {
-                if (!isNil(this.snippet.gistOwnerId)) {
-                    return this.snippet.gistOwnerId === profile.login;
+                if (isNil(this.snippet.gist)) {
```
WrathOfZombies: Curious. Can you link me to the implement of `isNil`. Would `_.isEmpty` not work?

src/client/app/effects/github.ts
Zlatkovsky: Not sure if that's what Dan meant, but I think you should throw your own instance of an error.
Looks like we do already have one, actually:  `export class PlaygroundError extends CustomError {`.  Please use that one instead.  Any exception is an instance of `Error` so as is, your `if` check doesn't do much.

Zlatkovsky:
@cshashwat , could you do the couple remaining small changes, and then let me know and I can merge in the changes?  thanks.

src/client/app/containers/app.ts
cshashwat: https://lodash.com/docs/#isNil

`isNil` covers both null and undefined. `isEmpty` sees if "value is an empty object, collection, map, or set", and strings specifically are considered empty with a length of 0. Not sure how it would handle undefined, which is what this condition is looking for.

src/client/app/effects/github.ts
cshashwat: Replaced with `PlaygroundError`

src/client/app/effects/github.ts
```diff
@@ -115,7 +115,7 @@ export class GitHubEffects {
             .map((gist: IGist) => ({ gist: gist, snippetId: id }))
             .catch(exception => {
                 if (!gistOwnerId && exception.status >= 400 && exception.status <= 499) {
-                    throw new Error(JSON.stringify({ id: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
+                    throw new PlaygroundError(JSON.stringify({ snippetId: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
```
Zlatkovsky: Should you add some sort of "type: UpdateGistFailed"?  So that you can then use it in your `.catch` block to check that this is exactly what you wanted (and not some other PlaygroundError that we eventually might add somewhere, which might be thrown from upstream?)

src/client/app/effects/github.ts
```diff
@@ -141,14 +141,13 @@ export class GitHubEffects {
                 new Snippet.UpdateInfoAction({ id: snippetId, gist: gist.id, gistOwnerId: gist.owner.login })])
         )
         .catch(exception => {
-            if (exception instanceof Error) {
-                exception = JSON.parse(exception.message);
+            let action = null;
+            if (exception instanceof PlaygroundError) {
+                let { snippetId, gistOwnerId } = JSON.parse(exception.message);
```
Zlatkovsky: Same as above, check the type to make sure that it's an error that you're supposed to act upon.

Also you should wrap this parsing code with a try/catch.  It would be a shame for the error-handler to throw an error if the input is, for whatever reason, invalid (i.e., it was a PlaygroundError but just with a text message, not the same one as you were throwing above)...

src/client/app/effects/github.ts
cshashwat: Added code to check this.

src/client/app/containers/app.ts
WrathOfZombies: Then `isEmpty` is a super set of `isNil`. I'd use that instead as if the gist is empty, you don't want to do anything anyways.

Zlatkovsky:
:shipit:

Issue #469, opened 2017-07-05
BLOCKER:  bornholm-edge deployment is broken (unless in in-private window)


Zlatkovsky:
@dsaunders45459 , @cshashwat , it looks like accessing the bornholm-edge site in IE 11 (browser or add-in, doesn't matter) is showing broken (missing) UI when not in in-private browser mode.  I am not sure if it's a matter of having snippets already, or of having cached files; but in either case, it's not working.  We need to make sure it doesn't make it in like that to insiders/prod (possibly by implementing the hashing & cache-busting first).

This is happening to me on both my home machine and laptop.  I am not going to flush my cache yet, so that I can remain a repro case.  Let me know if you'd like me to try something.

![image](https://user-images.githubusercontent.com/2230453/27848877-5ad7e136-60fd-11e7-9f3d-b2e22771c9d2.png)

Zlatkovsky:
Adding @WrathOfZombies too, in case he can offer some advice.  Just launching F12 didn't reveal much.  No console errors, though a suspiciously-empty initial root state...

![image](https://user-images.githubusercontent.com/2230453/27848925-c843fe3a-60fd-11e7-9106-aa2cfb2c1cd9.png)

versus the in-private one:

![image](https://user-images.githubusercontent.com/2230453/27848954-0ed8a0ee-60fe-11e7-86c0-f3ee574016e2.png)

Zlatkovsky:
I should also mention that this issue was happening even before I merged in #468.  So it's not the new string changes.

dsaunders45459:
I think this is probably related to an n-1 browser cache issue with the JavaScript resources, and not to localStorage. I can repro fairly often using the following steps on IE 11, although it doesn't repro on Chrome. The browsers have different cache policies when the response headers are not being set, so this is probably why I can only hit it in IE. @cshashwat can you look into the chunk hashing next and wait for it to deploy the changes to prod.

Clear IE cache
Fiddler "urlreplace bornholm-edge.azurewebsites.net/ script-lab.azureedge.net/"
Navigate to https://bornholm-edge.azurewebsites.net/ in IE
Click on Excel theme icon
Click on Basic API call
Save the snippet
Fiddler urlreplace clear
Close/reopen browser and renavigate to https://bornholm-edge.azurewebsites.net/ (may be optional)
ctrl + r to reload page. Most of the time the page will fail to load here or no-op after clicking Excel icon again.

PR #474, opened 2017-07-06
Add chunk hashing for js and css bundles

cshashwat:
## Description

The  JS and CSS bundles generated from webpack now have a generated hash added to their filenames, forcing the browser to refresh its files.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/469

## Motivation and Context

See above issue. The problem seems to be an issue with the loading of bundle resources (specifically the Javascript ones), which can be repro'd frequently by switching between domains via Fiddler. Adding a chunkhash to the filename should stop this issue and prevent the browser from loading a mix of old and new files, that may break the application.

## How Has This Been Tested?

Tested by repro'ing the error described above, implementing these changes and confirming that they solved the issue.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -1,8 +1,8 @@
 const webpack = require('webpack');
 const path = require('path');
+const AssetsWebpackPlugin = require('assets-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
-const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const { CheckerPlugin } = require('awesome-typescript-loader');
 const autoprefixer = require('autoprefixer');
 const perfectionist = require('perfectionist');
@@ -43,14 +43,6 @@ module.exports = (prodMode) =>
                     exclude: /node_modules/
                 },
                 {
-                    test: /\.scss$/,
-                    use: ExtractTextPlugin.extract({
-                        fallback: "style-loader",
-                        use: ['css-loader', 'postcss-loader', 'sass-loader']
-                    }),
-                    exclude: /theme/
-                },
-                {
                     test: /\.(png|jpe?g|gif|svg|woff|woff2|ttf|eot|ico)$/,
                     use: {
                         loader: 'file-loader',
@@ -83,8 +75,11 @@ module.exports = (prodMode) =>
                     }
                 }
             }),
+            new AssetsWebpackPlugin({
+                filename: 'assets.json',
+                path: path.resolve('./dist/server')
+            }),
             new CheckerPlugin(),
-            new ExtractTextPlugin('[name].[chunkhash].bundle.css'),
             new webpack.optimize.CommonsChunkPlugin({
                 name: ['vendor', 'polyfills'],
                 minChunks: Infinity
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -84,7 +84,7 @@ module.exports = (prodMode) =>
                 }
             }),
             new CheckerPlugin(),
-            new ExtractTextPlugin('[name].bundle.css'),
+            new ExtractTextPlugin('[name].[chunkhash].bundle.css'),
             new webpack.optimize.CommonsChunkPlugin({
                 name: ['vendor', 'polyfills'],
                 minChunks: Infinity
--- a/src/server/interfaces.ts
+++ b/src/server/interfaces.ts
@@ -20,7 +20,7 @@ export interface IRunnerHandlebarsContext {
 
     origin: string;
     host: string;
-    assets: { [key: string]: string };
+    assets: { [key: string]: any };
 
     initialLoadSubtitle: string;
     headerTitle: string;
@@ -37,7 +37,7 @@ export interface IRunnerHandlebarsContext {
 
 export interface IErrorHandlebarsContext {
     origin: string;
-    assets: { [key: string]: string };
+    assets: { [key: string]: any };
     title: string;
     message: string;
     details: string;
--- a/src/server/interfaces.ts
+++ b/src/server/interfaces.ts
@@ -20,7 +20,7 @@ export interface IRunnerHandlebarsContext {
 
     origin: string;
     host: string;
-    assets: any;
+    assets: { [key: string]: string };
 
     initialLoadSubtitle: string;
     headerTitle: string;
@@ -37,7 +37,7 @@ export interface IRunnerHandlebarsContext {
 
 export interface IErrorHandlebarsContext {
     origin: string;
-    assets: any;
+    assets: { [key: string]: string };
     title: string;
     message: string;
     details: string;
--- a/src/server/interfaces.ts
+++ b/src/server/interfaces.ts
@@ -20,6 +20,7 @@ export interface IRunnerHandlebarsContext {
 
     origin: string;
     host: string;
+    assets: any;
 
     initialLoadSubtitle: string;
     headerTitle: string;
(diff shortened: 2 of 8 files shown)


WrathOfZombies:
@cshashwat @Zlatkovsky We didnt do the chunk hash due to the fact that we were referencing the js/css files manually in some html files that are not generated by webpack. If all our html are being generated by the `HtmlWebpackPlugin` then I am ok with this.

config/webpack.common.js
```diff
@@ -84,7 +84,7 @@ module.exports = (prodMode) =>
                 }
             }),
             new CheckerPlugin(),
-            new ExtractTextPlugin('[name].bundle.css'),
+            new ExtractTextPlugin('[name].[chunkhash].bundle.css'),
```
WrathOfZombies: Move this out of common and have separate variants for development and production. `chunkhash` reduces incremental build perf when used in development

WrathOfZombies:
If you want hashing you need to consider generating templates dynamically and have travis deploy them to the server instead of checking them in. 

Observe the following as examples. No way of predicting what the hash was:
- https://github.com/OfficeDev/script-lab/blob/master/src/server/templates/error#L11
- https://github.com/OfficeDev/script-lab/blob/master/src/server/templates/error#L12
- https://github.com/OfficeDev/script-lab/blob/master/src/server/templates/error#L29

src/server/server.ts
```diff
@@ -47,6 +47,15 @@ function isOfficeHost(host: string) {
     return officeHosts.indexOf(host) >= 0;
 }
 
+let assetPaths = null;
```
Zlatkovsky: You generally don't need to set to null.  Just leave it as its default undefined value, it's ok.

src/server/server.ts
```diff
@@ -47,6 +47,15 @@ function isOfficeHost(host: string) {
...
+function getAssetPaths() {
+    if (assetPaths === null) {
```
Zlatkovsky: >assetPaths === null [](start = 8, length = 19)

could probably just do "!assetPaths".  Since assetPaths is expected to be an object (i.e., empty string or "false" are not valid values for it in any case), I think just doing "!assetPaths" is fine.  Or check for null *or* undefined (if you leave it as undefined, per above)

src/server/server.ts
```diff
@@ -104,6 +114,7 @@ registerRoute('get', '/run/:host/:id', (req, res) => {
                 returnUrl: '',
                 origin: currentConfig.editorUrl,
                 host: host,
+                assets: assets,
```
Zlatkovsky: >assets [](start = 24, length = 6)

why don't you just inline it?

    assets: getAssetPaths();

src/server/server.ts
```diff
@@ -47,6 +47,15 @@ function isOfficeHost(host: string) {
...
 
+let assetPaths = null;
+function getAssetPaths() {
+    if (assetPaths === null) {
+        let data = fs.readFileSync(path.resolve(__dirname, 'assets.json'));
+        assetPaths = JSON.parse(data.toString());
+    }
+
+    return assetPaths;
+}
```
Zlatkovsky: Do you need it to be a function, or could we just always initialize the paths from the very start.  I don't see why not.  Basically, what I mean is:

    const assetPaths = getAssetPaths();

And where getAssetPaths is a singly-called helper that does

    const data = fs.readFileSync(path.resolve(__dirname, 'assets.json'));
    return JSON.parse(data.toString());

src/server/server.ts
Zlatkovsky: Though per one of the comments above, you could just make it a global and not need to evaluate the function at all.

---
In reply to: [126536827](https://github.com/OfficeDev/script-lab/pull/474#discussion_r126536827) [](ancestors = 126536827)

src/server/interfaces.ts
```diff
@@ -20,6 +20,7 @@ export interface IRunnerHandlebarsContext {
 
     origin: string;
     host: string;
+    assets: any;
```
Zlatkovsky: >ssets: any; [](start = 5, length = 11)

Is there not a better type than any?  For example, it sounds like it's just a key-value pair JSON object.  If so, maybe declare it as `{ [key: string]: string }`?  Ditto inside the `server.ts` code as well.

config/webpack.common.js
```diff
@@ -83,8 +75,11 @@ module.exports = (prodMode) =>
                     }
                 }
             }),
+            new AssetsWebpackPlugin({
+                filename: 'assets.json',
+                path: path.resolve('./dist/server')
+            }),
```
Zlatkovsky: Could you paste in what the output looks like in a comment (just here on the code review, so I can try to imagine what it would work like)?  Thanks!

Zlatkovsky:
:clock1:

src/server/server.ts
cshashwat: Fixed.

src/server/server.ts
cshashwat: Changed it to "!assetPaths".

src/server/server.ts
cshashwat: Inlined the function, like you suggested, so there's no need for the extra variable declaration.

src/server/server.ts
cshashwat: Logic behind having it as a function was that if, for some reason, the file read fails, then reloading the page would call the function again, triggering a re-read. Having a `const` or something similar at the top would not accomplish that, I believe. But am open to other suggestions.

src/server/interfaces.ts
cshashwat: Added.

config/webpack.common.js
cshashwat: Added as comment below.

cshashwat:
The generated `assets.json` file from the `AssetsWebpack` plugin, is written to `dist/server` and has sample output like so

`
{"vendor":{"js":"vendor.bundle.js"},"main":{"js":"main.bundle.js","css":"main.bundle.css"},
"runner":{"js":"runner.bundle.js","css":"runner.bundle.css"},
"gallery":{"js":"gallery.bundle.js","css":"gallery.bundle.css"},
"error":{"js":"error.bundle.js","css":"error.bundle.css"},........
`

src/server/server.ts
dsaunders45459: Yes if the file read intermittently fails when the server starts up the static could be stuck in a bad state until iis reset.

src/server/interfaces.ts
Zlatkovsky: Actually, wait, it looks like it's not a string-to-string but rather a string-to-object map.  In that case it should be { [key: string]: any }

---
In reply to: [126537988](https://github.com/OfficeDev/script-lab/pull/474#discussion_r126537988) [](ancestors = 126537988)

src/server/server.ts
Zlatkovsky: OK, fair enough.

---
In reply to: [126569799](https://github.com/OfficeDev/script-lab/pull/474#discussion_r126569799) [](ancestors = 126569799)

Zlatkovsky:
:+1:

config/webpack.common.js
```diff
@@ -1,8 +1,8 @@
 const webpack = require('webpack');
 const path = require('path');
+const AssetsWebpackPlugin = require('assets-webpack-plugin');
```
dsaunders45459: Do you need to add this to yarn.lock?

cshashwat: From looking at the documentation, `yarn.lock` is a auto-generated file and shouldn't be edited.

src/server/interfaces.ts
cshashwat: Fixed.

Issue #497, opened 2017-07-12
One "beta" ribbon is enough


jakobpn:
I think we are overdoing it with 2 ribbons. I suggest we simplify and only show the ribbon to the right.

![image](https://user-images.githubusercontent.com/13109307/28131842-03a7baaa-66f0-11e7-931a-8c9047ea007d.png)

cshashwat:
@Zlatkovsky what's your opinion on this? The justification for two was that the user's eyes might automatically go to one side of the screen vs the other, but in the default add-in pane, it seems like one might be enough.

Issue #495, opened 2017-07-12
Use sentence casing in Share menu


jakobpn:
![image](https://user-images.githubusercontent.com/13109307/28128094-767aa5e0-66e3-11e7-9d9e-ab9d7b8b5322.png)

Change "New Public Gist" to "New public gist"
Change "New Secret Gist" to "New secret gist"
Change "Update Existing Gist" to "Update existing gist"

All other menu items are using sentence casing.

Zlatkovsky:
@cshashwat , is this one closed now?  Please close the issues once the environment that they are targeting has received the update.

I also like to use the "resolved-awaiting-prod-deployment" label, at least for prod ones.  Please close any issues marked with that, once you and @dsaunders45459 push to production next week.

Issue #494, opened 2017-07-12
Unlocalized word in the switch environment dialogue


maninweb:
Hi there,

from the About dialogue, I can now switch the environment to another one, that's cool. However, the title of the environment switcher dialogue includes the unlocalized string `to`.

![image](https://user-images.githubusercontent.com/13045898/28110568-1dea29a4-66f3-11e7-951c-1f2dc586ac67.png)

It can be found in `about.ts` within the code block starting with `let changeEnvironmentResult`. Would it be possible to also put the string `to` into the language files? Or, as a suggestion, have something parametrized like "Switch from _VAR_1 to VAR_2"? Thanks :-)

Best,
Mourad

Zlatkovsky:
Thanks for the find.  @cshashwat , please add the string and then, as with the others, re-open the #472 issue making note of the new string, so that Mourad can translate it.  Thanks!

maninweb:
Hi, for me this issue is solved within the last Insider Version, thanks :-)
I did not close this issue, as I prefer someone of you do this.

Zlatkovsky:
Closing.  By the way, we had intended to push to production this week, but found an issue that needs to be resolved first.  So, Dan and Shashwat will work on getting production updated early next week.

Issue #492, opened 2017-07-12
Colors of "Alpha" and "Beta" banners


Zlatkovsky:
Couple issues:

1. The red color for "beta" looks scary (and jumps out much more than "alpha", even though it's alpha that should be the scary one!)
2. The colors don't look "office"-y.

Could you:
1. Use one of the colors in the Office pallet?  Take one of the hex values out of one of the themes (e.g., the PPT orange color, or the Word/Outlook blue color).
2. Reverse the colors such that alpha has the orange-ish one that draws attention, whereas beta has the more informational-esque blue, for example.

Thanks!

![image](https://user-images.githubusercontent.com/2230453/28101766-7c87623e-667f-11e7-9200-2bf9381ba9e7.png)

PR #499, opened 2017-07-13
Address insiders bug bash issues

cshashwat:
## Description

This pull request addresses outstanding issues found in the recent insiders push:

- Ribbon colors have been modified to use the Office UI pallet color scheme
- Ribbon now only appears on right side
- Dropdown for environment switching now uses default styling, same as language switcher
- Previously unlocalized strings have been localized
- String casing has been changed to sentence case
- Sharing logic has been changed: if a user shares a gist for the first time, he/she will not be prompted with a confirmation dialog. However, afterwards, they will be, to get their confirmation that they want to create a brand new snippet. Strings have been updated accordingly. Strings have also been updated for the 'Update' action; the dialog box is now slightly different versus sharing a gist.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/497
https://github.com/OfficeDev/script-lab/issues/495
https://github.com/OfficeDev/script-lab/issues/494
https://github.com/OfficeDev/script-lab/issues/492
https://github.com/OfficeDev/script-lab/issues/491
https://github.com/OfficeDev/script-lab/issues/489

## Types of changes

## Checklist:

### DIFF
--- a/src/client/app/components/about.ts
+++ b/src/client/app/components/about.ts
@@ -63,9 +63,9 @@ export class About implements AfterViewInit {
     originalLanguage = '';
 
     configs = [
-        { name: 'Production', value: 'production' },
-        { name: 'Beta', value: 'insiders' },
-        { name: 'Alpha', value: 'edge' },
+        { name: this.strings.production, value: 'production' },
+        { name: this.strings.beta, value: 'insiders' },
+        { name: this.strings.alpha, value: 'edge' },
     ];
     selectedConfig = '';
 
--- a/src/client/app/components/about.ts
+++ b/src/client/app/components/about.ts
@@ -101,9 +101,9 @@ export class About implements AfterViewInit {
             return;
         }
 
-        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment;
-        changeEnvironmentMessage = changeEnvironmentMessage.replace('{0}', this.configs.find(c => c.value === currentConfigName).name);
-        changeEnvironmentMessage = changeEnvironmentMessage.replace('{1}', this.configs.find(c => c.value === this.selectedConfig).name);
+        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment.replace(
+            '{0}', this.configs.find(c => c.value === currentConfigName).name).replace(
+            '{1}', this.configs.find(c => c.value === this.selectedConfig).name);
         let changeEnvironmentResult = await this._effects.alert(
             this.strings.changeEnvironmentConfirm,
             changeEnvironmentMessage,
--- a/src/client/app/components/about.ts
+++ b/src/client/app/components/about.ts
@@ -101,9 +101,12 @@ export class About implements AfterViewInit {
             return;
         }
 
+        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment;
+        changeEnvironmentMessage = changeEnvironmentMessage.replace('{0}', this.configs.find(c => c.value === currentConfigName).name);
+        changeEnvironmentMessage = changeEnvironmentMessage.replace('{1}', this.configs.find(c => c.value === this.selectedConfig).name);
         let changeEnvironmentResult = await this._effects.alert(
             this.strings.changeEnvironmentConfirm,
-            `${this.strings.aboutSwitchEnvironment} ${this.configs.find(c => c.value === currentConfigName).name } to ${this.configs.find(c => c.value === this.selectedConfig).name }`,
+            changeEnvironmentMessage,
             this.strings.okButtonLabel,
             this.strings.cancelButtonLabel
         );
--- a/src/client/app/components/about.ts
+++ b/src/client/app/components/about.ts
@@ -101,9 +101,9 @@ export class About implements AfterViewInit {
             return;
         }
 
-        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment.replace(
-            '{0}', this.configs.find(c => c.value === currentConfigName).name).replace(
-            '{1}', this.configs.find(c => c.value === this.selectedConfig).name);
+        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment
+            .replace('{0}', this.configs.find(c => c.value === currentConfigName).name)
+            .replace('{1}', this.configs.find(c => c.value === this.selectedConfig).name);
         let changeEnvironmentResult = await this._effects.alert(
             this.strings.changeEnvironmentConfirm,
             changeEnvironmentMessage,
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -186,28 +186,27 @@ export class AppComponent {
                 }
 
                 let { isPublic, isUpdate } = values;
+                let confirmationAlertIfAny = null;
                 if (isUpdate) {
                     this._store.dispatch(new GitHub.UpdateGistAction(this.snippet));
                 }
                 else if (isPublic) {
                     if (this.snippet.gist) {
-                        let result = await this._effects.alert(this.strings.sharePublicSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel);
-                        if (result === this.strings.cancelButtonLabel) {
-                            return;
-                        }
+                        confirmationAlertIfAny = await this._effects.alert(this.strings.sharePublicSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel);
                     }
 
-                    this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
+                    if (confirmationAlertIfAny !== this.strings.cancelButtonLabel) {
+                        this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
+                    }
                 }
                 else {
                     if (this.snippet.gist) {
-                        let result = await this._effects.alert(this.strings.sharePrivateSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel);
-                        if (result === this.strings.cancelButtonLabel) {
-                            return;
-                        }
+                        confirmationAlertIfAny = await this._effects.alert(this.strings.sharePrivateSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel);
                     }
 
-                    this._store.dispatch(new GitHub.SharePrivateGistAction(this.snippet));
+                    if (confirmationAlertIfAny !== this.strings.cancelButtonLabel) {
+                        this._store.dispatch(new GitHub.SharePrivateGistAction(this.snippet));
+                    }
                 }
 
                 if (sub && !sub.closed) {
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -179,7 +179,7 @@ export class AppComponent {
         }
 
         let sub = this.isLoggedIn$
-            .subscribe(isLoggedIn => {
+            .subscribe(async (isLoggedIn) => {
                 if (!isLoggedIn) {
                     this._store.dispatch(new GitHub.LoginAction());
                     return;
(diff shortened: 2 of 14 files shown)


src/client/app/strings/german.ts
```diff
@@ -114,8 +116,8 @@ export function getGermanStrings(): ClientStrings {
         shareMenuPublic: 'Neues öffentliches GitHub-Gist',
         shareMenuPrivate: 'Neues geheimes GitHub-Gist',
         updateMenu: 'Bestehendes GitHub-Gist aktualisieren',
-        sharePublicSnippetConfirm: 'Sind Sie sicher, dass dieses Schnipsel in einem neuen öffentlichen GitHub-Gist geteilt werden soll?',
-        sharePrivateSnippetConfirm: 'Sind Sie sicher, dass dieses Schnipsel in einem neuen geheimen GitHub-Gist geteilt werden soll?',
+        sharePublicSnippetConfirm: englishSubstitutesForNotYetTranslated.sharePublicSnippetConfirm,
```
Zlatkovsky: Instead of deleting the old, can you comment them out, and add a comment for Mourad to know why he needs to update them (and also what the old text was, and what changed)?

src/client/app/containers/app.ts
```diff
@@ -190,22 +190,24 @@ export class AppComponent {
...
                 }
                 else if (isPublic) {
-                    this._effects.alert(this.strings.sharePublicSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel)
-                        .then((result: string) => {
-                            if (result !== this.strings.cancelButtonLabel) {
-                                this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
-                            }
+                    if (this.snippet.gist) {
+                        let result = await this._effects.alert(this.strings.sharePublicSnippetConfirm, `${this.strings.share} ${this.snippet.name}`, this.strings.share, this.strings.cancelButtonLabel);
+                        if (result === this.strings.cancelButtonLabel) {
```
Zlatkovsky: Isn't this "`if (result === this.strings.cancelButtonLabel) {`" statement shared between both public and private (and there is nothing else other than those two?)?  Should you move it out to outside the if statement?

src/client/app/components/about.ts
```diff
@@ -101,9 +101,12 @@ export class About implements AfterViewInit {
             return;
         }
 
+        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment;
+        changeEnvironmentMessage = changeEnvironmentMessage.replace('{0}', this.configs.find(c => c.value === currentConfigName).name);
+        changeEnvironmentMessage = changeEnvironmentMessage.replace('{1}', this.configs.find(c => c.value === this.selectedConfig).name);
```
dsaunders45459: You can write this as:
let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment.replace(
    '{0}', this.configs.find(c => c.value === currentConfigName).name).replace(
    '{1}', this.configs.find(c => c.value === this.selectedConfig).name);

src/client/app/strings/english.ts
```diff
@@ -75,6 +75,8 @@ export function getEnglishStrings(): ClientStrings {
         gistSharedDialogEnd: `To import your snippet, select the Import button in ${playgroundName} and enter this URL.`,
         gistSharedDialogTitle: 'Share your snippet',
         gistSharedDialogViewButton: 'View on GitHub',
+        gistUpdatedDialog: /** NEEDS STRING REVIEW */ 'The URL of your updated gist is the same as before:',
```
Zlatkovsky: I think the name of the string should more closely correspond to the title of the text.  I.e., `gistUpdateUrlIsSameAsBefore`

src/client/app/strings/german.ts
```diff
@@ -96,7 +98,7 @@ export function getGermanStrings(): ClientStrings {
         aboutSnippets: 'Lokale Schnipsel',
         aboutIntellisense: 'IntelliSense',
         aboutCurrentEnvironment: 'Aktuelle Umgebung:',
-        aboutSwitchEnvironment: 'Wechseln:',
+        aboutSwitchEnvironment: englishSubstitutesForNotYetTranslated.aboutSwitchEnvironment,
```
Zlatkovsky: See comment below re. the other strings that you cleared out.

src/client/app/strings/german.ts
```diff
@@ -4,8 +4,8 @@ import { ClientStrings } from './index';
 // Whenever these two lines are not commented out, it means that there are
 // still strings that need to be localized.
 // Just search for "englishSubstitutesForNotYetTranslated" in this file
-// import { getEnglishStrings } from './english';
-// const englishSubstitutesForNotYetTranslated = getEnglishStrings();
+import { getEnglishStrings } from './english';
+const englishSubstitutesForNotYetTranslated = getEnglishStrings();
```
Zlatkovsky: Could you re-open the #472 issue and make note that a few strings are being added (and a few others edited and need tweaking?)

src/client/app/effects/github.ts
```diff
@@ -113,22 +113,23 @@ export class GitHubEffects {
...
                 if (!gistOwnerId && exception.status >= 400 && exception.status <= 499) {
-                    throw new PlaygroundError(JSON.stringify({ type: 'UpdateGistFailed', snippetId: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
+                    throw new PlaygroundError(JSON.stringify({type: type, snippetId: id, gistOwnerId: UNKNOWN_GIST_OWNER_ID }));
                 }
                 throw exception;
             });
         })
-        .mergeMap(async ({ gist, snippetId }) => {
+        .mergeMap(async ({ type, gist, snippetId }) => {
             let temp = `https://gist.github.com/${gist.owner.login}/${gist.id}`;
```
dsaunders45459: can we rename temp to a better name such as gistUrl

dsaunders45459: Can you rename temp to gistUrl while you are in this file.

src/client/app/effects/github.ts
```diff
@@ -146,7 +147,7 @@ export class GitHubEffects {
             if (exception instanceof PlaygroundError) {
                 try {
                     let message = JSON.parse(exception.message);
-                    if (message.hasOwnProperty('type') && message.type === 'UpdateGistFailed') {
+                    if (message.hasOwnProperty('type') && message.type === GitHub.GitHubActionTypes.UPDATE_GIST) {
```
dsaunders45459: Is hasOwnProperty check really necessary? I would think you could just use if (message.type === GitHub.GitHubActionTypes.UPDATE_GIST)

src/client/public/index.script.ts
```diff
@@ -15,29 +15,27 @@ let { config } = PLAYGROUND;
...
-                break;
-        }
-
-        if (environment.current.config.name !== config['production'].name) {
-            ribbon.style.visibility = 'visible';
-        }
+    let ribbon = document.getElementById('ribbon');
+    switch (environment.current.config.name) {
+        // Utilize Office pallet colors for ribbons
+        case config['insiders'].name:
```
dsaunders45459: I think it is better for the compiler and intellisense if you use config.insiders.name instead config["insiders"].name for all of these cases

src/client/public/run.script.ts
```diff
@@ -2,28 +2,26 @@ import { environment } from '../app/helpers';
...
-                ribbon.style.background = 'green';
-                break;
-            default:
-                break;
-        }
+    let ribbon = document.getElementById('ribbon');
+    switch (environment.current.config.name) {
+        // Utilize Office pallet colors for ribbons
+        case config['insiders'].name:
+            ribbon.textContent = 'Beta';
```
dsaunders45459: Double check if we need to localize Alpha or Beta.

src/client/public/index.script.ts
```diff
@@ -15,29 +15,27 @@ let { config } = PLAYGROUND;
...
-                ribbon.style.background = 'green';
-                break;
-            default:
-                break;
-        }
-
-        if (environment.current.config.name !== config['production'].name) {
-            ribbon.style.visibility = 'visible';
-        }
+    let ribbon = document.getElementById('ribbon');
```
dsaunders45459: Seems like there is a lot of duplication between index and run for the ribbon, did we want to move it to it's own script?

src/client/app/strings/german.ts
cshashwat: Added.

src/client/app/containers/app.ts
cshashwat: I thought it made more sense to keep it inside the if statement, since the alert is only called if the gist is not new, and that's the only time we care about the user's selection.

src/client/app/strings/english.ts
cshashwat: Added.

src/client/app/effects/github.ts

src/client/app/effects/github.ts

src/client/public/index.script.ts
cshashwat: The compiler complains if I use dot notation. Not sure why, but intellisense doesn't recognize `config.insiders`

src/client/app/components/about.ts
cshashwat: Added.

src/client/app/containers/app.ts
Zlatkovsky: But it's shared between both statements.  So I think you can just have a `let confirmationAlertIfAny` defined above, set that variable with the `await this._effects.alert(...)`, and do the if check in one common spot below.

src/client/public/index.script.ts
Zlatkovsky: you can do `(config as any).insiders.name`

src/client/app/strings/german.ts
cshashwat: Done.

src/client/public/run.script.ts
cshashwat: @Zlatkovsky Is this an issue?

src/client/app/components/about.ts
```diff
@@ -101,9 +101,9 @@ export class About implements AfterViewInit {
             return;
         }
 
-        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment;
-        changeEnvironmentMessage = changeEnvironmentMessage.replace('{0}', this.configs.find(c => c.value === currentConfigName).name);
-        changeEnvironmentMessage = changeEnvironmentMessage.replace('{1}', this.configs.find(c => c.value === this.selectedConfig).name);
+        let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment.replace(
```
Zlatkovsky: Nit: my personal preference is actually to have the line break happen before the ".replace".

    let changeEnvironmentMessage = this.strings.aboutSwitchEnvironment
        .replace('{0}', this.configs.find(c => c.value === currentConfigName).name)
        .replace({1}', this.configs.find(c => c.value === this.selectedConfig).name);

Just you would with a `.then` or `.catch`

src/client/public/index.script.ts
cshashwat: Would be pretty straightforward. @Zlatkovsky should I add this, e.g. as a `ribbon.script.ts`?

src/client/app/containers/app.ts

src/client/app/components/about.ts
cshashwat: Added.

src/client/public/index.script.ts
Zlatkovsky: Sure, though I would be careful about the word "ribbon" (we generally mean the office ribbon tabs by "ribbon").  And what if we want to have other splash screen logic?  How about "splash-screen-common.script.ts".  Or dots instead of hypthens in that name, see what most other files do.

src/client/public/run.script.ts
Zlatkovsky: Not sure, but it won't hurt to add the texts to strings, I suppose.

src/client/app/helpers/splash.screen.ts
```diff
@@ -0,0 +1,30 @@
+import { environment } from './environment';
+import { Strings } from '../strings';
+let { config } = PLAYGROUND;
+
+let strings = Strings();
+
+export function instantiateRibbon() {
```
Zlatkovsky: I wonder if the id of the element (or the dom element itself) should be passed in (just for future flexibility's sake; what if I need to call the element something else?)

Zlatkovsky:
@cshashwat , see the one comment I just added.  And beyond that, merge in when ready.  I signed off as approved.

Issue #500, opened 2017-07-13
Insider version does not load localized strings for the ribbon (?)


maninweb:
Hi there,

I tested the Insider-Version and I noticed that the menue entries in the ribbon are not loaded in German, but in English in my German Excel 2016 Insider Slow Version 1706 Build 8229.2073 . The task panes are loading the localized strings correctly.

![image](https://user-images.githubusercontent.com/13045898/28160338-333c3522-67c0-11e7-9cf4-04f4de8a9eb5.png)
 
In the Script Lab About dialogue, German and Beta have been automatically set as language and as current environment. I did the following steps to load & test the Insider-Version:

* From my local sideload folder, I removed all manifests related to Script Lab and copied the last file version of `script-lab-insiders.xml` there.
* I refreshed the sideloaded Add-Ins within Excel (from Insert/My Addins/Shared folder) and added the Add-In. It then welcomed me in English.
* No change, if I restart Excel, for example.

Best,
Mourad

maninweb:
Hi,

for playing, I changed the country code in the `script-lab-insiders.xml` from `Locale="de"` to `Locale="de-de"`. This worked and the strings in the ribbon are in German now.

I then ran the local dev version and I have the same problem, that in the current dev version the ribbon menue entries appear in English. If I change the locale from `de` to `de-de` in the dev manifest and refresh my Add-Ins, then it works in the dev version again and the menue entries are in German. In my previous versions, `de` worked fine for me, while translating.

I will not exclude, that something happened to my Excel version, as while doing this above and opening Excel several times, sometimes Excel moved to the Add-In section from the Insert tab to the backstage and reverted this on a next run. Same in Word and PowerPoint too for me.

Best,
Mourad

Zlatkovsky:
Not sure if it's perhaps a sideloading on the network share issue?  @saunders77 , could you please try it out / investigate?

maninweb:
Hi,

for your info, I think, I have some bigger problems with my Office install since today. Besides the move of the store to the backstage (and back), the icons on the home tab (only) for Word and PowerPoint (but not in Excel, One Note and Outlook) are all suddenly gone and I see an empty space. Appeared after I switched Office from German to English and back and back again, if I remember well.

![image](https://user-images.githubusercontent.com/13045898/28174748-a71d4a6c-67f2-11e7-8a6f-af30caa0deb3.png)

[...]

PR #501, opened 2017-07-13
Translation to German for the meta data for the store

maninweb:
## Description
Translation to German for the meta data for the store.

## Related Issue
N/A

## Motivation and Context
So that it appears on the German version of the store.

## How Has This Been Tested?
N/A

3 screenshots in German for use in the store.

## Types of changes

- [X] Translation
## Checklist:

- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.

## Notes
The files include the version for `de-de` (Germany). I've seen that the store is also available for `de-at` (Austria) and `de-ch` (Switzerland). In case a version for Austria and Switzerland would have to be addressed by their own in the store, the files for Germany can also be used, as the translation is the same for those countries.


Zlatkovsky:
@saunders77 , FYI that Mourad has made the changes.  Merging now.  Btw, did you add to the translation doc the part about needing the description and images?

Zlatkovsky:
Also @saunders77 , once I merge this in, do you want to rename the folder (and your en-us one) to just "en" and "de", to correspond with the store language entries?  If we do start specializing the locales, we can then rename just the specialized ones (e.g., "en-gb")

While you're renaming folders, I would also use a hyphen in "store-metadata" instead of an underscore.  That is the way of the web...

saunders77:
@Zlatkovsky :

Yes, I added the instructions to the translation doc.

@locale-specific paths: that's the standard way Office works and it's the standard way the internet works: languages are locale-specific.

@store-metadata: what about our top-level node_modules folder? It would look weird to have:
`/store-metadata/`
`/node_modules/`

Zlatkovsky:
@saunders77 , for locale-specific paths:  yes, you can use locale in addition to language, but it's also ok to have just language (which essentially is the logical equivalent of "en-*").  Per the StackOverflow thread (https://stackoverflow.com/questions/44964786/localizing-manifest-files-for-an-office-js-add-in), the manifest can also do just the language, not locale.  Though perhaps, if that answer is incorrect, that explains the issue https://github.com/OfficeDev/script-lab/issues/500.  I wonder if it would work if it _was_ labeled as "de-de", for example.  Could you look into that issue?

For the folder name.  OK, fair enough.  I will admit that I always found "node_modules" naming awkward, but it's true that it is there.  Fine, we can keep it.

Issue #490, opened 2017-07-12
Menu doesn't hide when get the alert dialog, leading to a slightly strange visual experience


Zlatkovsky:
The menu will disappear when I hover away from it fully, but initially it looks interactive and as if it still has focus.

![image](https://user-images.githubusercontent.com/2230453/28100892-cf6ab448-6679-11e7-8941-7fcf1ba44daa.png)

![image](https://user-images.githubusercontent.com/2230453/28189950-61ed8728-67dd-11e7-84cd-9b16bc663ab2.png)

PR #504, opened 2017-07-14
Fix menu hide issue

cshashwat:
## Description

Share menu would not hide when the alert dialog appeared and would still be interactive, leading to potentially undefined user actions. This has now been fixed via adding a subscription to the app's sharing state, and manually removing the menu when the state is changed through CSS styling.

This has been implemented for the relevant Share Menu items, including Update, Public gist, Secret gist, and Copy to clipboard.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/490

## Types of changes

## Checklist:

### DIFF
--- a/src/client/app/containers/app.ts
+++ b/src/client/app/containers/app.ts
@@ -19,7 +19,7 @@ import { isEmpty } from 'lodash';
                     <command icon="Play" title="{{strings.runInThisPane}}" [async]="running$|async" (click)="run()"></command>
                     <command icon="OpenPaneMirrored" title="{{strings.runSideBySide}}" (click)="runSideBySide()"></command>
                 </command>
-                <command [hidden]="isEmpty" icon="Share" [async]="sharing$|async" title="{{strings.share}}">
+                <command [ngClass]="{'disabled': isDisabled}" [hidden]="isEmpty" icon="Share" [async]="sharing$|async" title="{{strings.share}}">
                     <command *ngIf="isGistOwned|async" icon="Save" title="{{strings.updateMenu}}" (click)="shareGist({isPublic: false, isUpdate: true})"></command>
                     <command icon="PageCheckedin" title="{{strings.shareMenuPublic}}" (click)="shareGist({isPublic: true, isUpdate: false})"></command>
                     <command icon="ProtectedDocument" title="{{strings.shareMenuPrivate}}" (click)="shareGist({isPublic: false, isUpdate: false})"></command>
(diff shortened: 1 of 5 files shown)


src/client/app/containers/app.ts
```diff
@@ -62,6 +63,10 @@ export class AppComponent {
             this.snippet = snippet;
         });
 
+        this._store.select(fromRoot.getSharing).subscribe(sharing => {
+            sharing ? this.isDisabled = true : this.isDisabled = false;
```
dsaunders45459: A better way to write this is "this.isDisabled = sharing"

src/client/app/containers/app.ts
```diff
@@ -198,6 +205,8 @@ export class AppComponent {
                     if (confirmationAlertIfAny !== this.strings.cancelButtonLabel) {
                         this._store.dispatch(new GitHub.SharePublicGistAction(this.snippet));
                     }
+
+                    this.isDisabled = false;
```
dsaunders45459: Is setting this.isDisabled necessary anywhere other than in the this._store.select(fromRoot.getSharing).subscribe observer?

src/client/app/effects/github.ts
```diff
@@ -173,12 +173,17 @@ export class GitHubEffects {
             AI.trackEvent(GitHub.GitHubActionTypes.SHARE_COPY, { id: rawSnippet.id });
             new clipboard('#CopyToClipboard', {
                 text: () => {
-                    this._uiEffects.alert(Strings().snippetCopiedConfirmation, null, Strings().okButtonLabel);
                     return yaml;
                 }
+            }).on('success', async() => {
+                await this._uiEffects.alert(Strings().snippetCopiedConfirmation, null, Strings().okButtonLabel);
+                this._store.dispatch(new GitHub.ShareSuccessAction(<IGist>{}));
```
dsaunders45459: Can we just pass in null here?

src/client/app/containers/app.ts
dsaunders45459: @Zlatkovsky what do you think, should we create a seperate store for dialog so that we can just disable the menu whenever a dialog is visible or should we just use this sharing store because it covers all the dialog/menu disabling scenarios today.

src/client/app/effects/github.ts

src/client/app/containers/app.ts
cshashwat: Yes, it's necessary here, because in the event that the user presses the 'Cancel' option, the sharing state will never get changed because no action is dispatched. So 'isDisabled' needs to be set to still allow the menu to show.

src/client/app/containers/app.ts

PR #515, opened 2017-07-17
Add fix for HTML caching issue: deploy old resources with new ones

cshashwat:
## Description

Chunk hashing was added for JS and CSS bundles, but due to the browser caching old HTML pages, old resources were referenced, causing '404' errors. This pull request solves this issue by deploying the new and old resources to the server. The pull request also creates a new 'bundles' directory at the root, to reference the generated files and avoid cluttering with other files.

## Types of changes

## Checklist:

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -104,8 +104,8 @@ function deployBuild(url, folder) {
         // Copy previously bundled client resources to deploy along with newly generated resources
         if (url === EDITOR_URL) {
             shell.exec('git clone ' + url + ' current_build');
-            shell.exec('cp -f current_build/*.js current_build/*.css .');
-            shell.exec('cp -f current_build/bundles/*.js current_build/bundles/*.css ./bundles');
+            shell.exec('cp -n current_build/*.js current_build/*.css .');
+            shell.exec('cp -n current_build/bundles/*.js current_build/bundles/*.css ./bundles');
             shell.exec('rm -rf current_build');
         }
         shell.exec('git init');
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -101,10 +101,11 @@ function deployBuild(url, folder) {
         let next_path = path.resolve(folder);
         shell.cd(next_path);
         const start = Date.now();
+        // Copy previously bundled client resources to deploy along with newly generated resources
         if (url === EDITOR_URL) {
             shell.exec('git clone ' + url + ' current_build');
-            shell.exec('\cp -f current_build/*.js current_build/*.css .');
-            shell.exec('\cp -f current_build/bundles/*.js current_build/bundles/*.css ./bundles');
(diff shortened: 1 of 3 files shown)


config/deploy.js
```diff
@@ -101,6 +101,12 @@ function deployBuild(url, folder) {
         let next_path = path.resolve(folder);
         shell.cd(next_path);
         const start = Date.now();
+        if (url === EDITOR_URL) {
+            shell.exec('git clone ' + url + ' current_build');
+            shell.exec('\cp -f current_build/*.js current_build/*.css .');
```
dsaunders45459: I don't think you need \ in front of the command, this just disables the alias

config/deploy.js
```diff
@@ -101,6 +101,12 @@ function deployBuild(url, folder) {
         let next_path = path.resolve(folder);
         shell.cd(next_path);
         const start = Date.now();
+        if (url === EDITOR_URL) {
```
dsaunders45459: Add a comment describing what you are doing here for future developers making changes to this file.

cshashwat: Added.

config/deploy.js
cshashwat: Had it because I saw online that 'cp' is often aliased to 'cp -i', but after checking, this does not seem to be an issue for us. Fixed.

config/webpack.dev.js
```diff
@@ -11,8 +11,8 @@ module.exports = () =>
 
         output: {
             path: path.resolve('./dist/client'),
-            filename: '[name].bundle.js',
-            chunkFilename: '[name].chunk.js',
+            filename: 'bundles/[name].bundle.js',
```
dsaunders45459: I know this is working, but I think bundles is part of the path and should not be in the filename. Can you look more into if we can add this to the path.

config/deploy.js
```diff
@@ -101,6 +101,13 @@ function deployBuild(url, folder) {
         let next_path = path.resolve(folder);
         shell.cd(next_path);
         const start = Date.now();
+        // Copy previously bundled client resources to deploy along with newly generated resources
+        if (url === EDITOR_URL) {
+            shell.exec('git clone ' + url + ' current_build');
+            shell.exec('cp -f current_build/*.js current_build/*.css .');
```
dsaunders45459: Actually it makes sense to use "cp -n" here instead of "cp -f". If WebPack for some reason has a name collision with previous file contents, we will prefer the newer version.

cshashwat: Added.

config/webpack.dev.js
cshashwat: Link to documentation indicating that using this notation is fine:

https://webpack.js.org/configuration/output/#output-filename

Issue #521, opened 2017-07-21
Runner is not setting no-cache headers


dsaunders45459:
Navigate to the runner error page:
https://bornholm-runner-edge.azurewebsites.net/run/excel2/3

We should use "no-cache, no-store" headers like we do for the editor server, but it looks like the web.config changes are not respected with the node.js server running. We may need to use express to set these headers.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser Name and version:
* Operating System and version (desktop or mobile):

## Useful logs

- [ ] Fiddler traces
- [ ] Console errors
- [ ] Screenshots
- [ ] Clip or GIFs

Issue #459, opened 2017-06-26
Set expiry date in response cache headers for Script Lab resources


dsaunders45459:
Running Google page speed test against Script lab shows some performance issues:
https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fscript-lab.azureedge.net

We need to be setting response cache headers for JavaScript/CSS resources, and then use chunk hash to force the browser to download the latest resource when they change. Without setting the expiration time the browser may request the resource from the server often, without the resource actually changing and making Script Lab slow to load.

Hongbo-Miao:
And turn on gzip can help a lot. My last project reduce from 5MB to around 200kb.

Zlatkovsky:
Related to https://github.com/OfficeDev/script-lab/issues/455

WrathOfZombies:
@Hongbo-Miao GZip is already enabled on the project. The headers are sent and the compression is working. What we do need is modifying the project build system to utilize AOT. That's going to give the biggest perf improvement. Could you help with that?

@dsaunders45459 given that these are served from a CDN, there's not much worry around the resource being requested again.

PR #522, opened 2017-07-24
Add server-side HTML no-cache headers and client-side static asset cache headers; modify deployment script to include static resources for up to 6 months

cshashwat:
## Description

Previously, server side HTML files did not have no-cache headers. This is now fixed.

Client-side files did have no-cache headers for HTML files but did not have cache headers for static resources; this is now fixed. Static resources are cached for up to 6 months.

To ensure safe deployment, the deploy script now deploys old resources along with new resources, if the old resources are less than 6 months old. After six months, these older resources are no longer deployed, and the browser should refresh.

## Related Issue

https://github.com/OfficeDev/script-lab/issues/521
https://github.com/OfficeDev/script-lab/issues/459

## Types of changes

## Checklist:

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node --harmony
 
 let path = require('path');
+let fs = require('fs');
 let chalk = require('chalk');
 let _ = require('lodash');
 let { build, config } = require('./env.config');
@@ -101,6 +102,9 @@ function deployBuild(url, folder) {
         let next_path = path.resolve(folder);
         shell.cd(next_path);
         const start = Date.now();
+        if (url === EDITOR_URL) {
+            buildAssetHistory(url);
+        }
         shell.exec('git init');
         shell.exec('git config --add user.name "Travis CI"');
         shell.exec('git config --add user.email "travis.ci@microsoft.com"');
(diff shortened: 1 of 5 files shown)


config/deploy.js
```diff
@@ -129,6 +133,42 @@ function deployBuild(url, folder) {
...
+    let newAssetsPath = path.resolve(__dirname, 'bundles');
+
+    // Parse old history file if it exists
+    let history = {};
+    try {
+        history = JSON.parse(fs.readFileSync(oldHistoryPath).toString());
+    } catch (e) {}
+
+    // Add new asset files to history, with current timestamp
+    let newAssets = fs.readdirSync(newAssetsPath);
```
dsaunders45459: Is this an absolute path, relative path, or filename returned here? We will want to use relative path to root.

config/deploy.js
```diff
@@ -129,6 +133,42 @@ function deployBuild(url, folder) {
...
+        history[asset] = { time: now };
+    }
+
+    try {
+        fs.accessSync(oldAssetsPath);
+        let existingAssets = fs.readdirSync(oldAssetsPath);
+        for (asset of existingAssets) {
+            let assetPath = path.resolve(newAssetsPath, asset);
+            // Check if old assets don't name-conflict and are less than six months old
+            if (!fs.existsSync(assetPath) && (now - history[asset].time < 15768000)) {
```
dsaunders45459: Should we delete the older entries from the json file as well through "delete history[asset]"

src/server/server.ts
```diff
@@ -317,6 +318,7 @@ function compileCommon(req: express.Request, res: express.Response, wrapWithRunn
             }
 
             timer.stop();
+            res.setHeader('Cache-Control', 'no-cache, no-store');
```
dsaunders45459: POST responses are not cacheable unless cache-control headers are set, so this may not be necessary.

dsaunders45459: You need to add cache-control headers to the error page as well.
    const html = await generateErrorHtml(error, strings);
    return res.contentType('text/html').status(200).send(html);

config/deploy.js
cshashwat: Discussed offline. These are just the filenames.

config/deploy.js
cshashwat: Discussed offline. For right now, have decided to keep the history file for all files, in the event that users start running into script lab errors, we could look up their files here.

src/server/server.ts
cshashwat: Added.

PR #535, opened 2017-07-28
Add view mode for Script-Lab: Users can now embed links to OfficeJS snippets and GitHub gists

cshashwat:
## Types of changes

## Checklist:


src/client/app/components/snippet.info.ts
```diff
@@ -43,6 +43,9 @@ export class SnippetInfo {
     @Output() dismiss = new EventEmitter<ISnippet>();
 
     strings = Strings();
+    constructor() {
+        console.log(this.snippet);
```
dsaunders45459: Remove this console log

src/client/app/components/view.ts
```diff
@@ -0,0 +1,72 @@
+import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+import { Store } from '@ngrx/store';
+import * as fromRoot from '../reducers';
+import { UI, Snippet } from '../actions';
+import { environment, applyTheme } from '../helpers';
+import { Strings } from '../strings';
+
+@Component({
+    selector: 'view',
```
dsaunders45459: can we come up with a better name for what this component's function is? view is too generic

src/client/app/components/view.ts
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        this._store.select(fromRoot.getCurrent).subscribe(snippet => {
+            this.snippet = snippet;
+        });
+    }
+
+    ngOnInit() {
+        let sub = this._route.params
+            .subscribe(async(params) => {
+                if (environment.current.host !== params.host.toUpperCase()) {
```
dsaunders45459: Can you use environment.current.host.toUpperCase() !== params.host.toUpperCase() instead

src/client/app/components/view.ts
```diff
@@ -0,0 +1,72 @@
...
+                    environment.current.host = params.host.toUpperCase();
+                    await applyTheme(environment.current.host);
+                }
+
+                if (/private-samples/.test(location.hash)) {
```
dsaunders45459: This should be included in params correct? There shouldn't be a need to directly test the hash.

dsaunders45459: We want to be able to easily switch to history API instead of hash based routing

src/client/app/containers/app.ts
```diff
@@ -10,40 +10,43 @@ import { isEmpty } from 'lodash';
...
-            <editor></editor>
-            <footer class="command__bar command__bar--condensed">
-                <command icon="Info" title="{{strings.about}}" (click)="showAbout=true"></command>
-                <command id="feedback" [title]="Feedback" icon="Emoji2" (click)="feedback()"></command>
-                <command icon="Color" [title]="theme$|async" (click)="changeTheme()"></command>
-                <command icon="StatusErrorFull" [title]="(errors$|async)?.length" (click)="showErrors()"></command>
-                <command class="language" [title]="language$|async"></command>
-            </footer>
-        </main>
+    <main *ngIf="!isViewMode" [ngClass]="theme$|async">
```
dsaunders45459: It is better if this goes into it's own component and makes use of angular default route to show. If this is too much code refactor, we can open a tracking issue to refactor. It seems unnecessary to have 2 *ngIf="!isViewMode", you can have a parent div to these two div that does the ngIf.

src/client/app/containers/app.ts
```diff
@@ -87,6 +90,10 @@ export class AppComponent {
             });
     }
 
+    get isViewMode() {
+        return /#\/view/.test(location.hash);
```
dsaunders45459: If we refactored the existing code into a new component, we wouldn't need this location.hash test and it would make it easier to use history API.

src/client/app/containers/editor.ts
```diff
@@ -39,7 +39,11 @@ export class Editor implements AfterViewInit {
      * Initialize the component and subscribe to all the necessary actions.
      */
     async ngAfterViewInit() {
-        this._monacoEditor = await this._monaco.create(this._editor, { theme: 'vs' });
+        let _overrides = { theme: 'vs' };
+        if (/#\/view/.test(location.hash)) {
```
dsaunders45459: This should be a parameter passed to Editor from "view" component and not looking at the location.hash.

src/client/app/components/view.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Strings } from '../strings';
+
+@Component({
+    selector: 'view',
+    template: `
+        <main [ngClass]="theme$|async">
+            <header class="command__bar">
+                <command class="title view-mode" [hidden]="isEmpty" icon="AppForOfficeLogo" [title]="snippet?.name"></command>
+            </header>
+            <editor readonly></editor>
```
dsaunders45459: you pass parameters to editor here, can you read "readonly"

src/client/assets/styles/components/_command.scss
```diff
@@ -31,6 +31,10 @@ command {
             }
         }
     }
+
+    &.view-mode {
+        pointer-events: none;
```
dsaunders45459: Why is this needed? Doesn't Monaco support readonly?

src/client/main.ts
```diff
@@ -14,16 +15,29 @@ import { SERVICE_PROVIDERS, MonacoService } from './app/services';
...
+    { path: '',   redirectTo: '/', pathMatch: 'full' }
+];
+
 let imports = [
     BrowserModule,
     HttpModule,
     FormsModule,
+    RouterModule.forRoot(
+      appRoutes,
+      { enableTracing: true, useHash: true } // <-- debugging purposes only
```
dsaunders45459: Make sure to remove tracing before merging pull request

src/client/app/components/snippet.info.ts
cshashwat: Removed.

src/client/app/components/view.ts
cshashwat: Changed this to 'view-mode'. The corresponding "normal" app is 'editor-mode'

src/client/app/components/view.ts
cshashwat: Added.

src/client/app/components/view.ts
cshashwat: This won't be included in params: in `main.ts` the 'gist' route has a different set of parameters than the 'private-samples' route, so the route includes 'gist'/'samples'/'private-samples' to differentiate between each route. So the only way to check is via testing the hash. If we switch to history API we can test the pathname.

src/client/app/containers/app.ts
cshashwat: Moved this code to its own component, called 'EditorMode'. The bootstrap 'AppComponent' is now empty except for containing a `router-outlet` to tell Angular where to put the appropriate component.

src/client/app/containers/app.ts
cshashwat: Done.

src/client/app/containers/editor.ts
cshashwat: Added this.

src/client/app/components/view.ts
cshashwat: Added.

src/client/assets/styles/components/_command.scss
cshashwat: This is for the title bar at the top, which is not part of the monaco-editor. Normally clicking on it would show a menu or something; I've disabled it so there are no effects, hover or clickwise, that show when the user interacts with the div.

src/client/main.ts
cshashwat: Yup, will remove this once pull request is approved.

src/client/app/containers/view.mode.ts
```diff
@@ -0,0 +1,74 @@
...
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, rawUrl));
+                } else if (/gist/.test(location.hash)) {
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.GIST_VIEW, params.id));
+                } else {
+                    let rawUrl = `${environment.current.config.samplesUrl}/samples/${params.host}/${params.segment}/${params.name}.yaml`;
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, rawUrl));
+                }
+
+            });
+        sub.unsubscribe();
```
dsaunders45459: I don't think we want to unsubscribe here, we want to do it in the ngOnDestroy. I read that the router may not recreate the view component when hash changes, so we will want to make sure the new parameters get loaded in the editor in this case.

src/client/app/components/view.ts
dsaunders45459: this._route.snapshot.url can get you the url segments without having to use the dependency on hash. Can you look into using this instead?

src/client/app/effects/snippet.ts
```diff
@@ -34,6 +34,10 @@ export class SnippetEffects {
                 .filter(({ snippet }) => !(snippet == null))
                 .mergeMap(({ snippet, mode }) => this._massageSnippet(snippet, mode))
                 .catch((exception: Error) => {
+                    if (/#\/view/.test(location.hash)) {
```
dsaunders45459: Prefer that we pass parameter here for isViewMode instead of testing the hash

src/client/app/containers/view.mode.ts
```diff
@@ -0,0 +1,85 @@
...
+            .subscribe(async(params) => {
+                if (environment.current.host.toUpperCase() !== params.host.toUpperCase()) {
+                    environment.current.host = params.host.toUpperCase();
+                    // Update environment in cache
+                    environment.current = environment.current;
+                    await applyTheme(environment.current.host);
+                }
+
+                let urlSegments = this._route.snapshot.url;
+                switch (urlSegments[1].path) {
```
dsaunders45459: Can you make sure the proper telemetry is in place for view URL, we need to be able to differentiate view mode vs editor import. We also need to know what they are viewing for public/private gist we can log the full id, for gist we need to log the hash value.

dsaunders45459:
Another request is accept a correlation id parameter for deep linking.

src/client/app/effects/snippet.ts
```diff
@@ -360,7 +365,7 @@ export class SnippetEffects {
          * If a imported snippet is a SAMPLE, then skip the save (simply to avoid clutter).
          * The snippet will get saved as soon as the user makes any changes.
          */
-        if (mode !== Snippet.ImportType.SAMPLE) {
+        if (mode !== Snippet.ImportType.SAMPLE && mode !== Snippet.ImportType.GIST_VIEW) {
```
dsaunders45459: If you add isViewMode parameter you probably do not need the GIST_VIEW enum.

src/client/assets/styles/components/_command.scss
dsaunders45459: Make sure that all the event handlers for this button are disabled as well when in view mode. When we enable keyboard accessibility we don't want to be able to invoke the action through keyboard instead of mouse.

src/client/views/test-view-iframe.html
```diff
@@ -0,0 +1,7 @@
+<html>
+
+<body class="ms-font-m">
+	<iframe src="https://localhost:3000/#/view/private-samples/excel/01-basics/basic-api-call" style="height:300px; width:600px;"></iframe>
```
dsaunders45459: This should not contain localhost it should be relative to the root. <iframe src="/#/view/private-samples/excel/01-basics/basic-api-call"

cshashwat: Added.

src/client/app/containers/view.mode.ts
cshashwat: Added this under `_massageSnippet()` in `effects/snippets.ts`

src/client/app/effects/snippet.ts
cshashwat: Added this as an optional parameter in the `Snippet.ImportAction`

src/client/app/effects/snippet.ts
```diff
@@ -356,11 +362,21 @@ export class SnippetEffects {
             new Snippet.ImportSuccessAction(snippet)
         ];
 
+        let properties = {};
+        if (mode === Snippet.ImportType.GIST) {
+            properties = { ...properties, hashedGistId: sha1(snippet.gist)};
+        }
+        if (mode === Snippet.ImportType.SAMPLE) {
```
dsaunders45459: This can be else if

src/client/app/effects/snippet.ts
```diff
@@ -356,11 +362,21 @@ export class SnippetEffects {
             new Snippet.ImportSuccessAction(snippet)
         ];
 
+        let properties = {};
+        if (mode === Snippet.ImportType.GIST) {
+            properties = { ...properties, hashedGistId: sha1(snippet.gist)};
```
dsaunders45459: I don't think properties needs to be immutable, you can just do properties['HashGistId'] = sha1(snippet.gist) it will be better performance.

src/client/app/containers/view.mode.ts
```diff
@@ -0,0 +1,85 @@
...
+    strings = Strings();
+    snippet: ISnippet;
+    showInfo: boolean;
+    paramsSub: Subscription;
+
+    constructor(
+        private _store: Store<fromRoot.State>,
+        private _route: ActivatedRoute
+    ) {
+        this._store.select(fromRoot.getCurrent).subscribe(snippet => {
```
dsaunders45459: Lets clean this up later because we do it in other places too, but we should always unsubscribe OnDestroy or this subscription will hold a reference to ViewMode and it will leak ViewMode object.

Issue #547, opened 2017-08-03
With clean localStorage English it not the selected default locale in combo box


dsaunders45459:
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser Name and version:
* Operating System and version (desktop or mobile):

## Useful logs

- [ ] Fiddler traces
- [ ] Console errors
- [ ] Screenshots
- [ ] Clip or GIFs

PR #548, opened 2017-08-03
Fix language selection

dsaunders45459:
https://github.com/OfficeDev/script-lab/issues/547

## Types of changes

## Checklist:


src/client/app/strings/index.ts
```diff
@@ -50,7 +50,17 @@ export function Strings(): ClientStrings {
...
+    if (rawDisplayLanguage) {
+        rawDisplayLanguage = rawDisplayLanguage.toLowerCase();
+        for (let language of availableLanguages) {
+            if (rawDisplayLanguage === language.value) {
+                return rawDisplayLanguage;
+            }
+        }
+    }
+
+    return (rawDisplayLanguage || 'en').substr(0, 2);
```
Zlatkovsky: Should we start supporting the "en-us" forms here as well?

src/client/app/strings/index.ts
```diff
@@ -50,7 +50,17 @@ export function Strings(): ClientStrings {
 }
 
 export function getDisplayLanguage() {
-    return (getRawDisplayLanguage() || 'en').toLowerCase();
+    let rawDisplayLanguage = getRawDisplayLanguage();
```
Zlatkovsky: Why not `let rawDisplayLanguage = (getRawDisplayLanguage() || 'en').toLowerCase()`?  And that way you can avoid the if statement and the separate "toLowerCase".  You can even make it `const`, and not need to re-assign the variable...

src/client/app/strings/index.ts
dsaunders45459: I think "en" is fine for now until we start adding other regions.

Zlatkovsky: But I thought that in the pull request to add Chinese, we now have some places where we do use the region/locale.  Is that not so?  It just feels like we're trimming out possibly-useful information, not for English itself (since we only have one at the moment), but for another language that we might get from the `getRawDisplayLanguage()` method.

Zlatkovsky: @dsaunders45459 , once you've taken a look at the above comment, feel free to check in as is if it's not a concern.  I wasn't sure if it's important or not in light of the new Chinese strings.

dsaunders45459: It is not a concern because we first check if the full language/region is translated, if so we choose that language because it exists in the list. If the region is not translated, we look up by primary language.

PR #556, opened 2017-08-08
View url refactor: shorten paths and switch between themes

cshashwat:
## Description

Office JS Snippets now include private and public samples, all of which are grouped in a `view` subdirectory by host-type. This change refactors the existing view url functionality to take into account these changes. The change also fixes the issue of not dynamically applying the host stylesheets when changing hosts; previously the stylesheets were only applied if they had not already been imported or on page load.

## Types of changes

## Checklist:

### DIFF
--- a/src/client/app/containers/view.mode.ts
+++ b/src/client/app/containers/view.mode.ts
@@ -75,7 +75,6 @@ export class ViewMode implements OnInit, OnDestroy {
                     this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.GIST, id, true));
                 } else {
                     // Redirect to error page
-                    window.localStorage.clear();
                     location.hash = '/view/error';
                 }
             });
--- a/src/client/app/containers/view.mode.ts
+++ b/src/client/app/containers/view.mode.ts
@@ -1,16 +1,18 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Store } from '@ngrx/store';
+import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
-import * as fromRoot from '../reducers';
 import { UI, Snippet } from '../actions';
-import { environment, applyTheme } from '../helpers';
+import { environment } from '../helpers';
+import * as fromRoot from '../reducers';
+import { Request, ResponseTypes } from '../services';
 import { Strings } from '../strings';
 
 @Component({
     selector: 'view-mode',
     template: `
-        <main [ngClass]="theme$|async">
+        <main ngClass="{{theme$|async}} {{host}}">
             <header class="command__bar">
                 <command class="view-mode" [hidden]="isEmpty" [title]="snippet?.name"></command>
             </header>
@@ -27,11 +29,11 @@ import { Strings } from '../strings';
 export class ViewMode implements OnInit, OnDestroy {
     strings = Strings();
     snippet: ISnippet;
-    showInfo: boolean;
     paramsSub: Subscription;
 
     constructor(
         private _store: Store<fromRoot.State>,
+        private _request: Request,
         private _route: ActivatedRoute
     ) {
         this._store.select(fromRoot.getCurrent).subscribe(snippet => {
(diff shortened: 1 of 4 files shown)


src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,54 @@ import { Strings } from '../strings';
...
+                    environment.current.host = host.toUpperCase();
                     // Update environment in cache
                     environment.current = environment.current;
-                    await applyTheme(environment.current.host);
                 }
 
-                let urlSegments = this._route.snapshot.url;
-                switch (urlSegments[1].path) {
+                switch (type) {
                     case 'private-samples':
```
Zlatkovsky: We no longer need "private-samples", right?

src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,54 @@ import { Strings } from '../strings';
...
                     case 'samples':
-                        let rawSamplesUrl = `${environment.current.config.samplesUrl}/samples/${params.host}/${params.segment}/${params.name}.yaml`;
-                        this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, rawSamplesUrl, true));
-                        break;
+                        let hostJsonFile = `${environment.current.config.samplesUrl}/view/${environment.current.host.toLowerCase()}.json`;
+                        return (this._request.get<JSON>(hostJsonFile, ResponseTypes.JSON)
+                            .map(data => ({ data: data, id: id }))
+                        );
+                    case 'gist':
+                        return Observable.of({ data: null, id: id});
```
Zlatkovsky: What is the "data" parameter?

src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,54 @@ import { Strings } from '../strings';
...
+                }
+            })
+            .subscribe(({ data, id }) => {
+                if (data && data[id]) {
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, data[id], true));
+                } else if (id) {
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.GIST, id, true));
+                } else {
+                    // Redirect to error page
+                    window.localStorage.clear();
```
Zlatkovsky: What is this with clearing local storage?!  Sounds quite terrible!!!  Won't that lose folks' snippets, if they were, say, on Office Online?

src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,54 @@ import { Strings } from '../strings';
...
+                    location.hash = '/view/error';
```
Zlatkovsky: What does that error page look like?

src/client/assets/styles/editor.scss
```diff
@@ -3,6 +3,7 @@
 @import 'fabric';
 @import 'components';
 @import 'containers';
+@import 'theme';
 
 .app {
     view-mode-error {
```
Zlatkovsky: Are "view-mode-error" and "view-mode"  angular components?  And so that's why they're not prefaced with a "." or a "#"?  Just want to make sure I understand this...

src/client/assets/styles/editor.scss
```diff
@@ -13,6 +14,36 @@
...
+            @include theme($project);
+        }
+
+        .teams {
+            @include theme($teams);
+        }
+
+        .web {
+            @include theme($web);
+        }
```
Zlatkovsky: As of yesterday, we now need to include "Outlook" too.

Is it possible to perhaps have this bit of code elsewhere or re-use an existing bit of similar logic, so we don't need to remember to do it every time?  If not, it's OK, but in that case we should at least have a comment in both this block and the other, saying that "be sure to change the other location as well when you add a new theme".

Or maybe we can pull out both bits of code into their own file, a "global-host-to-theme-mapping.scss" or whatever...

src/client/app/containers/view.mode.ts
cshashwat: We still need some kind of differentiation in the path, to distinguish between samples and gists. We could technically have just a 'sample' vs 'gist' (because we now have the all samples JSON file so only need the ID) but I kept the private-samples vs sample for the purposes of explicit differentiation.

src/client/app/containers/view.mode.ts
cshashwat: The data parameter here is the JSON playlist data for the host, containing the mappings of IDs for  to their raw urls.

src/client/assets/styles/editor.scss
cshashwat: Yes, that's the reason why they have no preface

src/client/app/containers/view.mode.ts
cshashwat: Here is a URL on edge to see the current error page:

https://bornholm-edge.azurewebsites.net/#/view/error

src/client/assets/styles/editor.scss
cshashwat: Can discuss with you more offline about the theme logic; the reason that we even need this code is that the existing `applyTheme()` function from `theme.ts` explicitly loads the CSS stylesheets through the `NodeRequire` module. As a result, when you try to load the same stylesheet twice (say you load a gist for excel then word and then back to excel) it'll keep the previous one and not re-load it (so the gist for excel would be loaded but with a word theme instead of the excel theme).

src/client/app/containers/view.mode.ts
cshashwat: Removed this line.

src/client/app/containers/view.mode.ts
Zlatkovsky: I still don't get it.  `private-samples` would still include public samples in it, and vice-versa.  Let's just remove it and keep it only to options we care about.

dsaunders45459: just rename to samples

src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,53 @@ import { Strings } from '../strings';
...
-                    case 'gist':
-                        this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.GIST, params.id, true));
-                        break;
                     case 'samples':
-                        let rawSamplesUrl = `${environment.current.config.samplesUrl}/samples/${params.host}/${params.segment}/${params.name}.yaml`;
-                        this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, rawSamplesUrl, true));
-                        break;
+                        let hostJsonFile = `${environment.current.config.samplesUrl}/view/${environment.current.host.toLowerCase()}.json`;
+                        return (this._request.get<JSON>(hostJsonFile, ResponseTypes.JSON)
+                            .map(data => ({ data: data, id: id }))
```
Zlatkovsky: snippetDictionary

src/client/app/containers/view.mode.ts
```diff
@@ -27,43 +29,53 @@ import { Strings } from '../strings';
...
+                    case 'gist':
+                        return Observable.of({ data: null, id: id});
                     default:
-                        break;
+                        return Observable.of({ data: null, id: null});
+                }
+            })
+            .subscribe(({ data, id }) => {
+                if (data && data[id]) {
+                    this._store.dispatch(new Snippet.ImportAction(Snippet.ImportType.SAMPLE, data[id], true));
```
Zlatkovsky: true /*isViewMode*/

cshashwat: Added.

src/client/app/containers/view.mode.ts
cshashwat: Added.

src/client/app/containers/view.mode.ts
cshashwat: Renamed.

PR #563, opened 2017-08-11
Prompt user before creating new gist in the event of a name conflict. Also fix bug where new files were created in GitHub after updating and changing name

cshashwat:
## Description

Previous Script Lab behavior when importing a snippet was, in the event of a name conflict with an already existing snippet, to automatically modify the name with a trailing number.

Users are now prompted when importing a snippet where there is a name conflict and have one of three options:
- 'Go to existing snippet', which cancels the import action and opens up the first snippet that has the same name
- 'New snippet', which creates the new snippet with the new name as before
- 'Cancel', which cancels the import action entirely and does nothing

Screenshots are provided below.

Another issue addressed in this pull request: previously if users changed the name of a gist and then tried to update it, it would create an entirely new file within the gist. This behavior has been fixed so that the file name is the only thing that changes.

## How Has This Been
 Tested?

capture1

## Types of changes

## Checklist:


src/client/app/effects/snippet.ts
```diff
@@ -34,6 +34,7 @@ export class SnippetEffects {
                 .map((snippet: ISnippet) => ({ snippet, mode }))
                 .filter(({ snippet }) => !(snippet == null))
                 .mergeMap(({ snippet, mode }) => this._massageSnippet(snippet, mode, isViewMode))
+                .mergeMap(actions => (actions))
```
dsaunders45459: Does this do anything?

cshashwat: massageSnippet now returns a Promise of an Observable (Promise<Observable<>>) because of the async/await code, which needs to be just an Observable for the purposes of the effect, hence the extra step.

src/client/app/effects/snippet.ts
```diff
@@ -372,15 +361,45 @@ export class SnippetEffects {
...
+            );
         }
 
-        return Observable.from(actions);
+        let actions: Action[] = [];
+        if (importResult === Strings().cancelButtonLabel) {
+            return Observable.from([]);
+        } else if (importResult === Strings().snippetImportExistingButtonLabel) {
+            storage.snippets.values().some(item => {
+                if (item.name.trim() === snippet.name.trim()) {
```
dsaunders45459: No need to do this comparison twice.
if (item.name.trim() === snippet.name.trim()) {
actions.push(new Snippet.ImportSuccessAction(item));
return true;
}

return false;

src/client/app/effects/snippet.ts
```diff
@@ -372,15 +361,45 @@ export class SnippetEffects {
...
+                    actions.push(new Snippet.ImportSuccessAction(item));
+                }
+                return item.name.trim() === snippet.name.trim();
+            });
+            return Observable.from(actions);
```
dsaunders45459: Why do you need this return in every block. After the if/else blocks why not just do a return Observable.from(actions);

src/client/app/effects/snippet.ts
dsaunders45459: Did you try:
 .mergeMap(({ snippet, mode }) => Observable
            .fromPromise(this._massageSnippet(snippet, mode, isViewMode)))

src/client/app/effects/snippet.ts
```diff
@@ -372,15 +361,45 @@ export class SnippetEffects {
...
+                Strings().snippetImportExistingButtonLabel, Strings().defaultSnippetTitle, Strings().cancelButtonLabel /* user options */
+            );
         }
 
-        return Observable.from(actions);
+        let actions: Action[] = [];
+        if (importResult === Strings().cancelButtonLabel) {
+            return Observable.from([]);
+        } else if (importResult === Strings().snippetImportExistingButtonLabel) {
+            storage.snippets.values().some(item => {
```
dsaunders45459: for(let item of storage.snippets.values())
{
break;
} is easier to understand, although they both work.

src/client/app/services/github.service.ts
```diff
@@ -119,10 +119,21 @@ export class GitHubService {
...
+                .flatMap(gist => {
+                    let oldFilename = Object.keys(gist.files)[0];
+                    let newFilename = Object.keys(files)[0];
+                    body.files = {};
+                    body.files[oldFilename] = {
+                        filename: newFilename,
+                        content: files[newFilename].content,
+                        language: files[newFilename].language
+                    };
+                    return this._request.patch<IGist>(url, body, ResponseTypes.JSON, this._headers);
```
dsaunders45459: This still works in the case where we do not rename the file correct?

src/client/app/services/github.service.ts
```diff
@@ -119,10 +119,21 @@ export class GitHubService {
         let url = `${this._baseUrl}/gists`;
         if (!(id == null)) {
             url += `/${id}`;
-            return this._request.patch<IGist>(url, body, ResponseTypes.JSON, this._headers);
+            return this.gist(id)
+                .flatMap(gist => {
```
dsaunders45459: flatMap is an alias for mergeMap, but we use mergeMap everywhere else. Probably best to keep the name consistent.

cshashwat: Fixed.

src/client/app/effects/snippet.ts
cshashwat: Added.

src/client/app/effects/snippet.ts
cshashwat: Added.

src/client/app/effects/snippet.ts
cshashwat: Changed this to the for of loop as discussed offline.

src/client/app/effects/snippet.ts
cshashwat: Discussed offline; this didn't work so sticking with the current solution for now.

src/client/app/effects/snippet.ts
```diff
@@ -372,14 +360,41 @@ export class SnippetEffects {
...
-         * The snippet will get saved as soon as the user makes any changes (if in editor mode).
+        /**
+         * If the action here involves true importing rather than re-opening,
+         * and if the name is already taken by a local snippet, ask the user if they want to:
+         * import the snippet with a new name, go to the already existing snippet, or cancel the import
          */
-        if (mode !== Snippet.ImportType.SAMPLE && !isViewMode) {
-            actions.push(new Snippet.SaveAction(snippet));
+        let importResult = null;
+        if (mode !== Snippet.ImportType.OPEN && this._exists(snippet.name)) {
```
Zlatkovsky: Whoa.  I don't think it's the name that we should be checking (names can change, that's what part of the other portion of your change is about).  And names, at least on a GitHub gist list, might well become non-unique ("Blank Snippet").

It's the GIST IDs that I care about, not their names!

src/client/app/effects/snippet.ts
```diff
@@ -372,14 +360,41 @@ export class SnippetEffects {
...
-         * If a imported snippet is a SAMPLE or the app is in view mode, then skip the save (simply to avoid clutter).
-         * The snippet will get saved as soon as the user makes any changes (if in editor mode).
+        /**
+         * If the action here involves true importing rather than re-opening,
+         * and if the name is already taken by a local snippet, ask the user if they want to:
+         * import the snippet with a new name, go to the already existing snippet, or cancel the import
          */
-        if (mode !== Snippet.ImportType.SAMPLE && !isViewMode) {
-            actions.push(new Snippet.SaveAction(snippet));
+        let importResult = null;
```
Zlatkovsky: could you specify the type?

src/client/app/effects/snippet.ts
```diff
@@ -372,14 +360,41 @@ export class SnippetEffects {
...
+                Strings().snippetNameDuplicationError,
+                `${Strings().importButtonLabel} ${snippet.name}`,
+                Strings().snippetImportExistingButtonLabel, Strings().defaultSnippetTitle, Strings().cancelButtonLabel /* user options */
+            );
         }
 
+        let actions: Action[] = [];
+        if (importResult === Strings().snippetImportExistingButtonLabel) {
+            for (let item of storage.snippets.values()) {
+                if (item.name.trim() === snippet.name.trim()) {
```
Zlatkovsky: again, gist IDs, not names

src/client/app/services/github.service.ts
```diff
@@ -119,10 +119,21 @@ export class GitHubService {
         let url = `${this._baseUrl}/gists`;
         if (!(id == null)) {
             url += `/${id}`;
-            return this._request.patch<IGist>(url, body, ResponseTypes.JSON, this._headers);
+            return this.gist(id)
+                .mergeMap(gist => {
+                    let oldFilename = Object.keys(gist.files)[0];
+                    let newFilename = Object.keys(files)[0];
+                    body.files = {};
```
Zlatkovsky: Feels weird to reset it here.  Should this not have been set on line 116 above, then?  And have body.files get set in the "else" clause?...

src/client/app/services/github.service.ts
```diff
@@ -119,10 +119,21 @@ export class GitHubService {
...
+                    body.files[oldFilename] = {
+                        filename: newFilename,
```
Zlatkovsky: Oh, interesting.  So even when we keep the same filename, you always use the rename syntax.  Ok, I guess that makes sense that GitHub would just no-op on this sort of stuff.  Though it would still be good to add a comment that this does a rename (or no-op).

src/client/app/strings/english.ts
```diff
@@ -40,6 +41,7 @@ export function getEnglishStrings(): ClientStrings {
         snippetImportErrorBody: `We couldn't import the snippet.`,
         reloadPrompt: 'Reload this task pane and then try another URL or ID.',
 
+        snippetNameDuplicationError: 'There is an existing snippet with this name. Do you want to go to this snippet or create a new one with a different name?',
```
Zlatkovsky: How about "There is an existing snippet that was imported from the same Gist ID.  Do you want to switch to this existing snippet, or create a new copy?"

Zlatkovsky:
Important note to whoever merges this:  remember to re-open string file bugs!

src/client/app/effects/snippet.ts
```diff
@@ -372,14 +360,41 @@ export class SnippetEffects {
...
+        /**
+         * If the action here involves true importing rather than re-opening,
+         * and if the name is already taken by a local snippet, ask the user if they want to:
+         * import the snippet with a new name, go to the already existing snippet, or cancel the import
          */
-        if (mode !== Snippet.ImportType.SAMPLE && !isViewMode) {
-            actions.push(new Snippet.SaveAction(snippet));
+        let importResult = null;
+        if (mode !== Snippet.ImportType.OPEN && this._exists(snippet.name)) {
+            importResult = await this._uiEffects.alert(
```
Zlatkovsky: Could we get this scenario (import existing already-in-workspace gist) into the markdown file that has the list of scenarios to test before deployment?

src/client/app/strings/english.ts
Zlatkovsky: Also, now that I think of it, I think we should do it for all GISTS, regardless of whether the user owns them or if he/she is importing someone else's.  Not sure if your current code covers that or not.

Could you add that to the manual test scenarios too?  (I.e., try to import the same external snippet twice.  Vs try to import your own snippet twice).

But note, only gists, not samples or anything else.  It should be totally reasonable for someone to import multiple copies of the same sample (esp. a "basic api sample").

src/client/app/effects/snippet.ts
cshashwat: This has been the existing logic since Script Lab's initial deployment; when we imported a snippet, we checked it against your existing snippets' names and then added the suffix if there was a name conflict. It made sense to me since otherwise your snippet list would grow confusing. We now want to change this logic entirely?

src/client/app/effects/snippet.ts
cshashwat: Added.

src/client/app/effects/snippet.ts
cshashwat: Discussed offline. Incorporated logic. Now if gist ID already exists among snippets, we notify the user. If the user wants a new snippet, we use the existing name logic to generate a new name.

src/client/app/services/github.service.ts
cshashwat: Added comment.

src/client/app/strings/english.ts
cshashwat: Added logic for this. Samples are not impacted, as intended.

src/client/app/effects/snippet.ts
cshashwat: Added.

src/client/app/effects/snippet.ts
```diff
@@ -187,6 +187,10 @@ export class SnippetEffects {
         .map((updatedSnippet) => new Snippet.SaveAction(updatedSnippet))
         .catch(exception => Observable.of(new UI.ReportErrorAction(Strings().snippetUpdateError, exception)));
 
+    private _gistIdExists(id: string) {
+        return storage.snippets.values().some(item => item.gist.trim() === id.trim());
+    }
+
     private _exists(name: string) {
```
Zlatkovsky: Nit: I might refactor this to "_nameExists" to make it even clearer.

src/client/app/services/github.service.ts
Zlatkovsky: I still think that we shouldn't set files in line 116.

src/client/app/effects/snippet.ts
```diff
@@ -34,9 +34,9 @@ export class SnippetEffects {
                 .map((snippet: ISnippet) => ({ snippet, mode }))
                 .filter(({ snippet }) => !(snippet == null))
                 .mergeMap(({ snippet, mode }) => this._massageSnippet(snippet, mode, isViewMode))
```
Zlatkovsky: I think "massageSnippet" is no longer the right name.  You're not massaging the snippet, you're instead importing it.  Maybe "_importHelper"?

src/client/app/effects/snippet.ts
```diff
@@ -187,6 +187,10 @@ export class SnippetEffects {
         .map((updatedSnippet) => new Snippet.SaveAction(updatedSnippet))
```
Zlatkovsky: Now that your "massageSnippet" does its own save too, do you need this one?  Or will it end up saving a sample, even though the sample shouldn't be saved?

PR #577, opened 2017-08-18
Support view mode import

cshashwat:


src/client/app/effects/snippet.ts
```diff
@@ -288,6 +288,13 @@ export class SnippetEffects {
             case Snippet.ImportType.SAMPLE:
             case Snippet.ImportType.URL:
             case Snippet.ImportType.GIST:
+                let viewUrlMatch = new RegExp(`^${environment.current.config.editorUrl}\/#\/view`);
+                if (type === Snippet.ImportType.URL && viewUrlMatch.test(data)) {
+                    /* If importing from a view-mode URL, simply redirect the user */
```
Zlatkovsky: I don't get it.  Don't we want to import rather than navigating the user away?..

src/client/app/effects/snippet.ts
```diff
@@ -288,6 +288,13 @@ export class SnippetEffects {
             case Snippet.ImportType.SAMPLE:
             case Snippet.ImportType.URL:
             case Snippet.ImportType.GIST:
+                let viewUrlMatch = new RegExp(`^${environment.current.config.editorUrl}\/#\/view`);
```
Zlatkovsky: I worry about taking semi-arbitrary strings and stuffing them into a regez.  What if they contain regex-special characters.  I would do a string match first instead, then trim out everything post the root URL, and regex only that.

Zlatkovsky: Also, should ignore capitalization (e.g., "VIEW")

dsaunders45459: It seems like there is a lodash method to escape a regular expression. We can probably just use this one:
_.escapeRegExp

src/client/app/effects/snippet.ts
cshashwat: By "import" do you mean actually importing the snippet and opening it in editor mode? The view mode already does an import action but doesn't save the snippet to local storage. We now want the user to be able to paste in a "view-url", but now they can edit that snippet? I thought that was accomplished by the open in playground feature.

cshashwat: Right now the view mode has access to Angular that parses the URL and gives it access to the param, such as host, gist vs sample, id, etc.. If we wanted to do it in the actual import feature, we would need our own regex matching and URL validation, in addition to the view mode code that already does this. Not sure if we want to do that?

Zlatkovsky: The "import" dialog in Script Lab should allow you to either paste in a GIST or a view URL pointing to a GIST -- and we should treat both as equally valid.

We can talk more in person if you want.

Zlatkovsky:
@cshashwat , ping on this one, just to get the pull request closed off...

Zlatkovsky:
BTW, we should support importing from any of the Script Lab environment URLs (e.g., localhost or prod should still be able to import a `https://bornholm-edge.azurewebsites.net/#/view/excel/sample/sdfasdfasdfasdf` snippet, and vice-versa)

Issue #593, opened 2017-08-30
[Regression] Heartbeat stops working after sharing a snippet


Zlatkovsky:
Not sure what the root cause is, but it seems like a 100% repro.  Seems to be for all hosts, though I repro-ed in web, since that has easiest debugging.

1. Start with a blank snippet in the web and a runner side-by-side
https://bornholm-edge.azurewebsites.net?mode=web
https://bornholm-edge.azurewebsites.net/run.html?gallery=true&mode=web

2. Create the snippet and then click on it in the runner.

So far so good, if you change code in the editor you should see the heartbeat-refresh message in the runner. (go ahead and refresh)

3. Now share the snippet as a secret gist.

4. Make a change in the editor.  No more heartbeat.  If you refresh the runner, you'll get the external-source-trust message (https://github.com/OfficeDev/script-lab/issues/592).  But even if you do trust, the heartbeat doesn't seem to come back.

BTW, not sure when exactly regressed, but it may have been a little while ago -- not sure if prod has this issue now as well.  However, it is from something in the last few weeks, it wasn't the case before.

WrathOfZombies:
Got this while following the repro steps:
![image](https://user-images.githubusercontent.com/3244360/29850861-6c06f8b6-8ce5-11e7-875d-b8975b80d7e9.png)

Zlatkovsky:
@WrathOfZombies,... interesting.  And possibly as worrysome, if not more.  You're getting this in the runner page?

Does this happen to you in an in-private window as well, or just in your current cache state?  what if you blow away the localSettings but not the cache, or vice-versa?

Filing it as a separate bug might be a good idea, just to make sure it doesn't get lost here with the heartbeat one...

WrathOfZombies:
Blowing the cache or incognito fixes the problem but I feel there's a bigger thing at play here:
![image](https://user-images.githubusercontent.com/3244360/29899458-80ea3a02-8da0-11e7-97f7-a95f1f6a54b8.png)

Looking at that image, it shows that we are trying to load the files from the runner domain and not the editor domain and that's cause of the fact that the paths begin with a `/` But I could swear that we print the `origin` before it, which means that the `origin` was set to `null` and I am not sure why that's happening.

Zlatkovsky:
Verified that the refresh works.  Closing.  (For Bhargav's other issue, it is being tracked separately -- though we need a repro)

Issue #592, opened 2017-08-30
After sharing a snippet, get trust prompt


Zlatkovsky:
Upon sharing a snippet and reloading, I get a trust prompt.

Repro:  start with blank snippet and runner side-by-side.  now share the snippet and reload runner, you'll see a trust prompt.

Whenever we share as a new Gist -- and provided the snippet was already trusted, i.e., wasn't a copy of someone else's imported gist that we never ran ourselves -- We should just always add a trust entry for the snippet.

Zlatkovsky:
Verified that it's addressed

Issue #591, opened 2017-08-30
"TrustedSnippets" list in storage grows indefinitely


Zlatkovsky:
And contains obsolete snippets, or -- I think -- even samples that I never ran.
We should probably use the editor-launch (where we load all the snippets in memory anyway) as a good time to quickly clean up that list, leaving only entries that are still part of localstorage + the currently lastOpened snippet.

Issue #585, opened 2017-08-26
No longer need import warning in editor


Zlatkovsky:
Now that we have the heartbeat-based warning (and let's check again that it works for all cases), no longer need this one:

![image](https://user-images.githubusercontent.com/2230453/29745264-f5053fb0-8a6a-11e7-88c3-452d86bdd9d3.png)

PR #595, opened 2017-08-30
Address heartbeat and trust snippet bugs, delete import warning

cshashwat:
## Description

This pull request addresses the following issues associated with the heartbeat and the current "Trust Snippet" dialog:

Trust Snippet list growing infinitely
Heartbeat not working upon sharing a snippet
After sharing a snippet, trust message still displays even though it should be trusted

## Related Issue

https://github.com/OfficeDev/script-lab/issues/593
https://github.com/OfficeDev/script-lab/issues/592
https://github.com/OfficeDev/script-lab/issues/591
https://github.com/OfficeDev/script-lab/issues/585

## Types of changes

## Checklist:


src/client/app/effects/github.ts
```diff
@@ -143,11 +143,17 @@ export class GitHubEffects {
 
                 return { gist: gist, snippetId: snippetId };
             })
-            .mergeMap(({ gist, snippetId }) => Observable.from([
+            .mergeMap(({ gist, snippetId }) => {
+                // Add snippet as trusted if it isn't already
+                if (!trustedSnippetManager.isSnippetTrusted(snippetId, gist.id)) {
```
Zlatkovsky: Realistically we're now creating a small security loophole -- that if you import someone's snippet, then share it, it's suddenly trusted, even if you never explicitly trusted the snippet in the first place.

I would make this conditional based on whether the snippet, before sharing, was one of the following:
* Was already trusted (in which case, no-op)
* Had a gist field that was empty
* Was a snippet that belongs to the user already (and hence is implicitly trusted)

src/client/app/effects/snippet.ts
```diff
@@ -61,6 +61,7 @@ export class SnippetEffects {
             const scrubbedSnippet = getScrubbedSnippet(rawSnippet, publicOrInternal);
             delete scrubbedSnippet.modified_at;
 
+            storage.snippets.load();
```
Zlatkovsky: Why do we need to load the entirety of the snippets?  What purpose does this code serve?

src/client/app/effects/snippet.ts
```diff
@@ -115,7 +116,10 @@ export class SnippetEffects {
     @Effect()
     loadSnippets$: Observable<Action> = this.actions$
         .ofType(Snippet.SnippetActionTypes.STORE_UPDATED, Snippet.SnippetActionTypes.LOAD_SNIPPETS)
-        .map(() => new Snippet.LoadSnippetsSuccessAction(storage.snippets.values()))
+        .map(() => {
+            trustedSnippetManager.reloadTrustedSnippets();
```
Zlatkovsky: is `reload` the right verb here?  Isn't it more like `cleanupTrustedSnippetsList`?

src/client/app/helpers/trusted.snippet.helper.ts
```diff
@@ -18,6 +20,18 @@ class TrustedSnippet {
         }
     }
 
+    reloadTrustedSnippets(): void {
+        try {
+            let trustedSnippets = JSON.parse(window.localStorage.getItem(this._trustedSnippetsKey));
+            for (let snippetId of Object.keys(trustedSnippets)) {
+                if (!storage.snippets.get(snippetId)) {
```
Zlatkovsky: This will only load the snippets list once?
Come to think of it, couldn't we just pass in the snippets list to the function outright as a parameter?

Zlatkovsky:
Also, are you sure it resolves https://github.com/OfficeDev/script-lab/issues/593?  I think that issue feels separate from the trust prompt, unless you can think of why they'd be related...

src/client/app/helpers/trusted.snippet.helper.ts
```diff
@@ -18,6 +20,18 @@ class TrustedSnippet {
         }
     }
 
+    reloadTrustedSnippets(): void {
```
Zlatkovsky: Probably fine to only call once per startup (e.g., in app.ts)

src/client/app/effects/snippet.ts
Zlatkovsky: // BUG number XYZ:  The load is just a workaround, look for a more performant solution later

WrathOfZombies: Can you also document the workaround here to keep track of things. Also please reference this PR in the bug.

src/client/app/effects/snippet.ts
WrathOfZombies: Honestly doesn't make a difference as you are rebuilding that list anyways.

src/client/app/helpers/trusted.snippet.helper.ts
```diff
@@ -1,3 +1,5 @@
+import { storage } from '../helpers';
+
 class TrustedSnippet {
     private _trustedSnippetsKey = 'trusted_snippets';
```
WrathOfZombies: would prefix this with `playground` check our other storage keys to see what's the prefix.

src/client/app/effects/github.ts
cshashwat: Added.

src/client/app/effects/snippet.ts
cshashwat: Added comment.

src/client/app/effects/snippet.ts
cshashwat: Changed name.

src/client/app/helpers/trusted.snippet.helper.ts
cshashwat: Added.

src/client/app/helpers/trusted.snippet.helper.ts
```diff
@@ -1,5 +1,19 @@
+import { storage } from '../helpers';
+
 class TrustedSnippet {
-    private _trustedSnippetsKey = 'trusted_snippets';
+    private _trustedSnippetsKey = 'playground_trusted_snippets';
```
dsaunders45459: this could just be a constant

PR #601, opened 2017-09-01
Add query string parameter for wacUrl, fix isSetSupported race condition

cshashwat:


src/client/public/runner.ts
```diff
@@ -47,6 +47,22 @@ interface InitializationParams {
 
     function initializeRunner(params: InitializationParams): void {
         try {
+            if (!((window.top as any).in_try_it_mode) && params.officeJS) {
+                let script = document.createElement('script');
+                script.src = params.officeJS;
+                script.addEventListener('load', (event) => {
```
dsaunders45459: if jQuery is already included on this page, it may be better to use $.getScript as it will work on older browsers. I am not sure this works on IE 9, but it is deprecated anyway so I am fine with this.

Zlatkovsky: We use `createElement` elsewhere (e.g., runner).  I think it's fine as is.

PR #611, opened 2017-09-09
Improve snippet-trust experience

Zlatkovsky:
In today's code, the snippet-trust was happening at the wrong point in time: adding to the trusted list when clicking "run".  This means that for snippets already shared before by the user, where there is already a gist ID, the trust experience would appear for his/her own snippets.

There were two other bugs:
* The trust dialog appeared above the others, so a snippet-switch message would not get show after you landed on an untrusted snippet.
* The trust dialog would sometimes get shown and then immeditely hidden, making the snippet appear frozen (and with no user action to unfreeze)


Zlatkovsky:
@dsaunders45459 , FYI, made a change to the trust code that Shashwat wrote.
I feel reasonably confident in the code, so am going ahead and just checking in as is.  But created it as a pull request for easier reviewability.
If you have comments, just let me know, and I can patch it up.

Issue #607, opened 2017-09-08
Getting a "permission denied" error when running snippet in the Code pane in Excel Online


jakobpn:
![image](https://user-images.githubusercontent.com/13109307/30190435-f3faae7e-93ee-11e7-8f45-b5b9614fcb5b.png)

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Browser Name and version:
* Operating System and version (desktop or mobile):

## Useful logs

- [ ] Fiddler traces
- [ ] Console errors
- [ ] Screenshots
- [ ] Clip or GIFs

Zlatkovsky:
Seems like a regression...

Zlatkovsky:
This was an issue (regression) that I introduced last week, but I believe that it's now fixed.  If not, please re-open.

PR #613, opened 2017-09-11
Fix broken runner in Office Online.

Zlatkovsky:
https://github.com/OfficeDev/script-lab/issues/607

Root cause: runner was calling "window.top", which on Office Online is in a different domain -- and hence was throwing an exception.

Fixed by surrounding logic with try/catch (and refactoring into separate function for cleanliness).  Also did a few other small refactorings.

BTW, changed window.top to window.parent, since ultimately someone could have nested the entire try-it experience in another website... and so it feels like window.parent is semantically more correct.

### DIFF
--- a/src/client/public/runner.ts
+++ b/src/client/public/runner.ts
@@ -44,9 +44,15 @@ interface InitializationParams {
         window: Window
     } = <any>{};
 
+    let isInTryItMode = checkIfInTryItMode();
+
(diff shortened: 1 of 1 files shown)


Zlatkovsky:
@dsaunders45459, created the pull request for easier diffing and record-keeping, but going ahead and merging as is.  If you have comments, just let me know...

Issue #639, opened 2017-09-23
Fix deployment issues for static lib files


Zlatkovsky:
Today, we correctly hash chunk files to avoid deployment issues wrt. caching on the browser.  Unfortunately, this is not applied to the "libs" folders that we copy (monaco, fabric, etc.)

With this change, start versioning those folders as well (based on their immutable package version number), and plumb that version-info throughout the rest of the system.

PR #640, opened 2017-09-23
Fix deployment issues for static lib files

Zlatkovsky:
Today, we correctly hash chunk files to avoid deployment issues wrt. caching on the browser. Unfortunately, this is not applied to the "libs" folders that we copy (monaco, fabric, etc.)

With this change, start versioning those folders as well (based on their immutable package version number), and plumb that version-info throughout the rest of the system.

#639


Zlatkovsky:
Seems to be working correctly locally.  Merging to alpha so I can see it work remotely, plus so that I can see the deploy logic run as well (for copying lib folders)

PR #654, opened 2017-09-29
Fix URL redirects (e.g., to alpha/beta)

Zlatkovsky:


config/env.config.js
```diff
@@ -85,15 +91,38 @@ class RedirectPlugin {
...
                                 return null;
-                            };
+                            }
+
+                            function isAllowedUrl(url) {
+                                if (url == null || url.trim().length === 0) {
+                                    return true;
+                                }
+                                
+                                url = decodeURIComponent(url);
```
dsaunders45459: Do we need to decodeURIComponent here, can't the caller be responsible for sending in a decoded URL?

config/env.config.js
```diff
@@ -85,15 +91,38 @@ class RedirectPlugin {
...
+
+                                var validRedirectLocations = ${JSON.stringify(validRedirectLocations)};
```
dsaunders45459: validRedirectLocations is already defined as const

config/env.config.js
```diff
@@ -85,15 +91,38 @@ class RedirectPlugin {
...
+
+                                var matchingLocation = validRedirectLocations.find(function(location) {
+                                    return location.indexOf(url) === 0;
```
dsaunders45459: We should probably do an exact match, not just looking for a substring as someone could register a new domain.

config/env.config.js
```diff
@@ -85,15 +91,38 @@ class RedirectPlugin {
...
+                                });
+
+                                return matchingLocation ? true : false;
```
dsaunders45459: !!matchingLocation

dsaunders45459: actually this will treat the 0 index as false, we need to test for null

config/env.config.js
```diff
@@ -126,9 +155,14 @@ class RedirectPlugin {
                             }
 
                             // If reached here, environment is already configured
-                        })();
+                            return;
+
+                        } catch (e) {
+                            console.log("Error redirecting the environments, staying on current page", e);
```
dsaunders45459: Probably use console.error instead

config/env.config.js
Zlatkovsky: We're effectively looking at starts-with (indexOf === 0).  I think that's better, because we don't want query parameters or etc to mess up the logic.  As long as we start with a fully-qualified domain that's in our whitelist, we're ok.

config/env.config.js
Zlatkovsky: Find returns the item, not the index.  So just the null (or undefined) is what we need to check for.
I thought about doing `!!matchingLocation`, but I think it's actually harder to understand.  I never knew about it until I saw Bhargav use it a year ago.  Is there a reason that `!!matchingLocation` is better than the `matchingLocation ? true : false` check that I'm doing?  I could be convinced either way

config/env.config.js
Zlatkovsky: Fair enough.  Really, it should be the `getParameterByName` that does it.  I'll make a change.

config/env.config.js
Zlatkovsky: Better yet, using Array.prototype.some:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

PR #663, opened 2017-10-15
Fix office.initialize not firing

Zlatkovsky:


Zlatkovsky:
@WrathOfZombies , @dsaunders45459, I made the fix (see commit description) to Office.initialize which was previously causing it to fire non-deterministically.

I am going ahead and merging this in into Insiders so that we can test it a bit, and then roll it into production (faster than it would go through master, which now has a bunch of other changes that will pend for a while).

If you have comments, please send them my way.

PR #748, opened 2018-02-16
ExcelMaker improvements

nico-bellante:
## Description

This pull request allows for maker scripts to be ran in the web as an Excel snippet. In addition, it adds functionality to allow for the maker scripts to be able to .getActiveWorkbook().


src/client/public/runner.ts
```diff
@@ -171,7 +187,9 @@ interface InitializationParams {
 
     /** Creates a snippet iframe and returns it (still hidden). Returns true on success
      * (e.g., snippet indeed shown, in contrast with, say, the Trust dialog being shown, but not the snippet) */
-    function replaceSnippetIframe(html: string, officeJS: string, isTrustedSnippet: boolean): boolean {
+    function replaceSnippetIframe(params: { html: string, officeJS: string, isTrustedSnippet: boolean, isMaker: boolean }): boolean {
```
Zlatkovsky: FYI, this DOES have a way of being done in TypeScript in a *somewhat* prettier way

    function replaceSnippetIframe(
        {html, officeJS, isTurstedSnippet, isMaker}
        : {html: string, officeJS: string, isTrustedSnippet: boolean, isMaker: boolean }
    ): boolean {
         ...
    }

src/runtime-helpers/maker.ts
```diff
@@ -2,33 +2,47 @@
...
+        }
+
         let _clientId: string;
+        let _scriptReferences: string[];
         let worker: Worker;
 
         export declare function getWorkbook(workbookUrl: string): Excel.Workbook;
+        export declare function getActiveWorkbook(): Excel.Workbook;
+
+        // todo figure out if this method can be hidden from intellisense
```
Zlatkovsky: Let's go an act on this TODO when we meet today.

src/runtime-helpers/maker.ts
```diff
@@ -47,48 +61,50 @@ module Experimental {
...
+                            if (['business', 'personal'].indexOf(driveType) < 0) {
+                                throw new Error('Currently, ScriptLab can only find urls for files stored on OneDrive consumer and business.');
+                            }
+
+                            let sliceIndex = driveType === 'business' ? 4 : 3;
+                            let path = tempUrl.split('/').slice(sliceIndex).join('/');
+                            let documentUrl = `${graphBaseUrl}/me/drive/root:/${path}:/workbook`;
+
+                            resolve(documentUrl);
+                        } else {
```
Zlatkovsky: Is there also an onfail sort of method on xhr, or does it always fire "onload" even on catostrophic error?

src/server/core/libraries.processor.ts
```diff
@@ -9,13 +9,17 @@
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 
-export function processLibraries(libraries: string) {
+export function processLibraries(libraries: string, isMakerScript: boolean, isInsideOffice: boolean) {
```
Zlatkovsky: Now that there are three parameters (of which two are booleans), it might be good to make this into an option bag.

src/server/maker/maker-worker.ts
```diff
@@ -1,6 +1,14 @@
...
 
+const VERBOSE_LOG = false;
+
+function ifVerbose(callback: () => void) {
+    if (VERBOSE_LOG) {
+        callback();
+    }
+}
+
 // TODO: import any script that the user feels like.
```
Zlatkovsky: Let's try to address this TODO

Zlatkovsky: Or is it already addressed?

src/server/maker/maker-worker.ts
```diff
@@ -99,27 +113,16 @@ module Experimental {
             }
 
             export function runMakerFunction(makerCode: string) {
-                let makerFunc: () => any;
+                let result: any;
 
                 // TODO EVENTUALLY: figure out if eval is evil.
```
Zlatkovsky: BTW, I think eval is not evil after all, in this context.  So probably OK to remove this todo.

src/server/server.ts
```diff
@@ -385,9 +388,9 @@ registerRoute('post', '/export', (req, res) => {
     // to avoid unhandled exception-pausing on debugging.
     return Promise.all([
         generateSnippetHtmlData(
-            { scriptToCompile: snippet.script, ...extractCommonCompileData(snippet) },
+            {scriptToCompile: snippet.script, ...extractCommonCompileData(snippet) },
             'script',
-            true /*isExternalExport*/, strings),
+            true /*isExternalExport*/, strings, true /*isInsideOffice*/),
```
Zlatkovsky: multiple boolean parameters that need explanation might be a code smell for making the function accept an options bag instead.

src/client/public/runner.ts
nico-bellante: is that really better? you have to type all the params twice..

PR #757, opened 2018-03-16
Removed waiting period when opening Script Lab in a browser tab

nico-bellante:
Removed waiting period when opening Script Lab in a browser tab

### DIFF
--- a/src/client/app/helpers/environment.ts
+++ b/src/client/app/helpers/environment.ts
@@ -287,10 +287,7 @@ class Environment {
             return new Promise<{ host: string, platform: string }>(async resolve => {
                 await this.createPlaygroundHostReadyTimer();
 
-                // Now wait for playground host-ready timer to return itself as ready.
-                // If it does, then clear off the previous timeout, and resolve immediately.
-
-                let { host, platform } = Utilities;
+                let { host } = Utilities;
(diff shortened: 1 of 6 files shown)


src/client/app/helpers/environment.ts
```diff
@@ -295,26 +285,23 @@ class Environment {
 
         const hostInfo = await (async (): Promise<{ host: string, platform: string }> => {
             return new Promise<{ host: string, platform: string }>(async resolve => {
-                let timer = this.createPlaygroundHostReadyTimer();
+                await this.createPlaygroundHostReadyTimer();
+
+                // Now wait for playground host-ready timer to return itself as ready.
```
Zlatkovsky: Is this a dead carry-over comment?

src/client/app/helpers/environment.ts
```diff
@@ -295,26 +285,23 @@ class Environment {
...
+                // If it does, then clear off the previous timeout, and resolve immediately.
 
-                let hostButtonsTimeout = setTimeout(() => {
+                let { host, platform } = Utilities;
```
Zlatkovsky: Now that I look at it, it feels like you could just do `const { host } = Utilities`.  And then do the platform bit later, when needed.

PR #761, opened 2018-03-23
bugfix - no refresh on desktop

nico-bellante:
* Sidestepped the storage helper to directly write and read from localStorage
* Setup polling in the heartbeat to check every 300ms whether a refresh is required

### DIFF
--- a/src/client/app/effects/snippet.ts
+++ b/src/client/app/effects/snippet.ts
@@ -14,7 +14,7 @@ import { Effect, Actions } from '@ngrx/effects';
 import * as cuid from 'cuid';
 import { Store } from '@ngrx/store';
 import * as fromRoot from '../reducers';
-import { isEmpty, isNil, find, assign, reduce, forIn, isEqual } from 'lodash';
+import { isEmpty, isNil, find, assign, reduce, forIn, isEqual, pick } from 'lodash';
 import * as sha1 from 'crypto-js/sha1';
 import { Utilities, HostType } from '@microsoft/office-js-helpers';
 const { localStorageKeys } = PLAYGROUND;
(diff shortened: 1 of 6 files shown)


src/client/app/effects/snippet.ts
```diff
@@ -96,8 +97,30 @@ export class SnippetEffects {
...
+
+            // update snippets
             scrubbedSnippet.modified_at = Date.now();
-            storage.snippets.insert(scrubbedSnippet.id, scrubbedSnippet);
+            let snippets = JSON.parse(window.localStorage.getItem(hostStorageKey));
+            snippets[scrubbedSnippet.id] = scrubbedSnippet;
+            window.localStorage.setItem(hostStorageKey, JSON.stringify(snippets));
+
+            // update lastOpened
+            let settings = JSON.parse(window.localStorage.getItem(localStorageKeys.settings));
```
Zlatkovsky: You could use lodash's `pick`:

`pick(object, ['a', 'c'])`

```
(() => {
    let { created_at, host, id, libraries, modified_at, name, description } = scrubbedSnippet;
    settings[environment.current.host].lastOpened = {
        created_at, host, id, libraries, modified_at, name, description
    };
})()
```

src/client/public/runner.ts
```diff
@@ -392,13 +384,12 @@ interface MakerInitializationParams {
...
                     $anotherSnippetSelected.hide();
                 } else {
-                    if (isListeningTo.snippetSwitching) {
-                        $anotherSnippetSelected.find('.ms-MessageBar-text .snippet-name').text(input.message.name);
-                        showReloadNotification($anotherSnippetSelected,
-                            () => clearAndRefresh(input.message.id, input.message.name, false /*isTrustedSnippet*/),
-                            () => isListeningTo.snippetSwitching = false,
-                            true /*allowShowLoadingDots*/);
-                    }
+                    // TODO: remove snippetSwitching if possible
```
Zlatkovsky: Let's test this on your machine.

Zlatkovsky: At minimum we should get rid of it entirely if we decide we truly don't need it.

PR #764, opened 2018-03-29
Bug fix local storage

nico-bellante:
## Types of changes

## Checklist:


src/client/app/effects/snippet.ts
```diff
@@ -103,7 +103,7 @@ export class SnippetEffects {
 
             // update snippets
             scrubbedSnippet.modified_at = Date.now();
-            let snippets = JSON.parse(window.localStorage.getItem(hostStorageKey));
+            let snippets = JSON.parse(window.localStorage.getItem(hostStorageKey)) || {};
```
Zlatkovsky: Will there ever be a case where the key doesn't even exist yet?  Should there be some sort of check for that too?

nico-bellante: Isn't that what this is?
`localStorage.getItem('keythatdoesntexist')` = null
and 
`JSON.parse(null)` = null

PR #774, opened 2018-04-11
Custom functions registration ui

nico-bellante:


src/server/core/snippet.helper.ts
```diff
@@ -119,3 +119,8 @@ export function getShareableYaml(rawSnippet: ISnippet, additionalFields: ISnippe
 export function isMakerScript(script: IContentLanguagePair) {
     return script.content.indexOf('Experimental.ExcelMaker') >= 0;
 }
+
+const isCustomFunctionRegex = /@customfunction/i;
+export function isCustomFunctionScript(script: IContentLanguagePair) {
```
nico-bellante: Oh, that friendly helper.  maybe you should use it elsewhere

src/server/interfaces.ts
```diff
@@ -63,14 +63,11 @@ export interface IRunnerHandlebarsContext {
 }
 
 export interface ICustomFunctionsRegisterHandlebarsContext {
-    snippets: ICustomFunctionsRegistrationRelevantData[];
-    snippetsDataBase64: string;
-
+    metadata: any;
```
nico-bellante: name something like "visualMetadata" and give it a type

src/client/app/helpers/custom.functions.helper.ts
```diff
@@ -1,30 +1,22 @@
...
-                catch {
-                    throw new Error(`Error parsing metadata for snippet "${snippet.name}"`);
-                }
-            });
+            .filter(snippet => snippet.script && isCustomFunctionRegex.test(snippet.script.content));
 
     let data: IRegisterCustomFunctionsPostData = {
         snippets: allSnippetsToRegisterWithPossibleDuplicate,
-        displayLanguage: getDisplayLanguage()
+        displayLanguage: getDisplayLanguage(),
```
Rickyp: For my understanding, why is the comma needed for the last property?

nico-bellante: If every line handles its own comma, then to delete a property, you can just delete the line. To add a property, you can just copy the line and modify the keys and values. 

I think (this)[https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8] summarizes it better.

nico-bellante: I also just like the uniform-ness of it ;P

src/client/app/containers/editor.mode.ts
```diff
@@ -137,11 +142,20 @@ export class EditorMode {
         }
     }
 
+    static _customFunctionsRegex = /@customfunction/i;
```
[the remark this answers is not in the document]
nico-bellante: yeah i forgot one

src/client/app/helpers/custom.functions.helper.ts
Rickyp: Sounds good to me!

PR #822, opened 2018-05-22
Prettier Formatting

nico-bellante:


Zlatkovsky:
@nico-bellante , would you like to merge (resolve conflict) and then commit this?

PR #837, opened 2018-06-19
Add streaming support for Custom Functions

nico-bellante:


src/runtime-helpers/custom-functions.ts
```diff
@@ -0,0 +1,4 @@
+interface IInvocationContext<T> {
```
jakobpn: I am wondering if we could make the declaration of a streaming custom function simpler by using a simpler name for this type. For example, ICaller<T>.

src/server/custom-functions/metadata.parser.ts
```diff
@@ -10,7 +10,7 @@ const TYPE_MAPPINGS = {
   [ts.SyntaxKind.BooleanKeyword]: 'boolean',
 };
 
-const CUSTOM_FUNCTION_OPTIONS_KEYS = ['stream', 'volatile'];
+const CUSTOM_FUNCTION_OPTIONS_KEYS = ['cancelable', 'volatile'];
```
jakobpn: Do we have support for "volatile" functions?

src/server/custom-functions/metadata.parser.ts
```diff
@@ -77,7 +85,22 @@ function traverseAST(sourceFile: ts.SourceFile): ICFFunctionMetadata[] {
...
-            if (func.type) {
+            if (isStreamingFunction) {
+              const lastParameterType = lastParameter.type as ts.TypeReferenceNode;
+              if (
+                !lastParameterType.typeArguments ||
+                lastParameterType.typeArguments.length !== 1
+              ) {
+                result = {
+                  error:
+                    'One and only one argument should be specified to IInvocationContext',
```
jakobpn: Is this error localized later?

src/__tests__/custom-functions/samples/streaming.txt
```diff
@@ -0,0 +1,118 @@
+// should pass streaming functions tests
+
+/**
+* increments a value each iteration
+* @param increment - number to add
+* @CustomFunction
+*/
+function incrementValue(increment: number, caller: IInvocationContext<number>): void {
```
Zlatkovsky: Please rename to `IInvocationContext` to `IStreamingCustomFunction`

src/__tests__/custom-functions/samples/streaming.txt
```diff
@@ -0,0 +1,118 @@
...
+         caller.setResult(result);
+    }, 1000);
+}
+
+/**
+* increments a value each iteration
+* @param increment - string to add
+* @CustomFunction
+*/
+function incrementValue(increment: number[][], caller: IInvocationContext<number[][]>) {
```
Zlatkovsky: What about adding a test for a FAILED streaming thing.  I.e., a caller that returns an object, or a Promise?

src/runtime-helpers/custom-functions.ts
Zlatkovsky: `IStreamingCustomFunction`
or
`IStreamingFunction`

src/server/custom-functions/metadata.parser.ts
Zlatkovsky: Good catch.  yeah, I don't think the platform does, actually.

Zlatkovsky: Actually, @nico-bellante , you might just want to remove both for now.  Either that, or add proper cancelable support (where "proper" means that if a function is declared as customable, I think it also gets the same interface as the streaming one but without setResult, just the oncancelled.. or something like that).  Honestly, I would just drop it for now.

src/server/custom-functions/metadata.parser.ts
```diff
@@ -57,15 +57,23 @@ function traverseAST(sourceFile: ts.SourceFile): ICFFunctionMetadata[] {
           if (isCF) {
             const jsDocParamInfo = getJSDocParams(func);
 
-            const parameters = func.parameters.map((p: ts.ParameterDeclaration) => {
-              const name = (p.name as ts.Identifier).text;
+            const [lastParameter] = func.parameters.slice(-1);
```
Zlatkovsky: What if there were 0 parameters?

Zlatkovsky: Also, style rules should force you to add spaces after `[` and before `]`

src/server/custom-functions/metadata.parser.ts
```diff
@@ -140,6 +167,15 @@ function traverseAST(sourceFile: ts.SourceFile): ICFFunctionMetadata[] {
 
 // helpers
 
+function checkLastParameterForStreaming(param: ts.ParameterDeclaration): boolean {
```
Zlatkovsky: How about `isLastParameterStreaming`

src/server/custom-functions/metadata.parser.ts
```diff
@@ -57,15 +57,23 @@ function traverseAST(sourceFile: ts.SourceFile): ICFFunctionMetadata[] {
...
 
-              return {
-                name,
-                ...(jsDocParamInfo[name] ? { description: jsDocParamInfo[name] } : {}),
-                ...getDimAndTypeHelper(p.type),
-              };
-            });
+            const parameters = func.parameters
+              .map((p: ts.ParameterDeclaration, i: number) => {
+                if (isStreamingFunction && i === func.parameters.length - 1) {
```
Zlatkovsky: This seems inelegant.  What about just doing a slice of func.parameters if if it a streaming function, skipping the last one?  That would be rather more functional, and cleaner..

src/server/custom-functions/metadata.parser.ts
nico-bellante: Agreed, but prettier doesn't seem to think so :/

Zlatkovsky:
@nico-bellante , please feel free to merge this to alpha (and also pull-request into beta, from there) once you're done with it...

PR #859, opened 2018-08-10
Master => Insider

Zlatkovsky:


Zlatkovsky:
Postponing till get a UI fix in Alpha

PR #888, opened 2018-12-13
Update to typescript 3.2.2

kylam-msft:
## Description

- Updated typescript to the latest version (3.2.2).
- Updated monaco-editor to the latest version (0.15.6) as well to support the updated version of typescript.

## Motivation and Context
TypeScript has introduced some cool features since 2.6.2 that would be neat to play around with or use in ScriptLab. Also, the latest developmental version of Office.js requires at least TS 3.0.

## How Has This Been Tested?

Local testing. Smoke testing with various snippets. Made sure existing code is not affected by any breaking changes (<https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes>).

## Types of changes

## Checklist:


Zlatkovsky:
Since we are now using the new Script Lab React codebase, closing this pull request as no-longer-needed.  The old Script Lab is now essentially in "escrow mode", we don't want to touch anything unless it's really broken (since it is now a backup option for folks using the new Script Lab, in case they need something that isn't working for them in the new one).

For the Monaco version bump, it might seems like an innocent change, but we actually found that there are very specific issue with Monaco across platforms for add-ins (namely, latest doesn't work with Mac add-ins!).  Which makes me all the more cautious about introducing new code to this repo.

